PR #14, opened 2017-10-23
BCN - Desiree & Sara M. 

saringa:
@dvillool

in progress

### DIFF
--- a/starter-code/main.js
+++ b/starter-code/main.js
@@ -1,9 +1,33 @@
+"use strict";
+
+var started = false;
+var seconds = 0;
+var idInterval;
+
+
 // Start/Stop Button
-document.getElementById("btnLeft").addEventListener("click", function () {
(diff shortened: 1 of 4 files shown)


starter-code/main.js
```diff
@@ -1,9 +1,33 @@
...
+var idInterval;
+
+
 // Start/Stop Button
-document.getElementById("btnLeft").addEventListener("click", function () {
+document.getElementById("btnLeft").addEventListener("click", function() {
   console.log("Left button clicked");
+
+  if (!started) {
+    seconds = 0;
```
tawebbcn: when you click "stop" and again click "start" you put seconds = 0.. means the timer rests. You should put seconds = 0 only when clicking the reset button.

tawebbcn:
The exercise was very difficult. have a look at the Examples here. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/prototype

PR #63, opened 2018-01-15
PAR Maxim / Josh 

joshvardey:


starter-code/javascript/chronometer.js
```diff
@@ -1,35 +1,43 @@
...
-
-// };
+function Chronometer() {
+  this.currentTime = 0;
+  this.intervalId = 0;
+}
+Chronometer.prototype.startClick = function() {
+  var that = this;
+  this.intervalId = setInterval(function() {
+    that.currentTime++;
```
ta-web-paris: Here you should also call the setTime method, so it's called on every interval

starter-code/javascript/chronometer.js
```diff
@@ -1,35 +1,43 @@
...
+};
 
 // Chronometer.prototype.setMilliseconds = function () {
 
 // };
 
-// Chronometer.prototype.stopClick = function () {
-  
-// };
+Chronometer.prototype.stopClick = function() {};
```
ta-web-paris: This method should user clearInterval to clear the interval you started on startClick

PR #64, opened 2018-01-15
PAR - Henri Delozanne

henridelozanne:


starter-code/javascript/chronometer.js
```diff
@@ -1,39 +1,49 @@
...
+function Chronometer() {
+  this.currentTime = 0;
+  this.intervalId = null;
+}
+
+Chronometer.prototype.startClick = function() {
+  var that = this;
+  that.intervalId = setInterval(function() {
+    that.currentTime += 1;
+    setTime();
```
ta-web-paris: This is failing because you need to use "that"
`that.setTime()`

PR #65, opened 2018-01-15
PAR - Jeff & Arthur M - Chronometer done

Qodibop:


starter-code/javascript/main.js
```diff
@@ -1,60 +1,43 @@
...
-
-}
+function setResetBtn() {}
 
 // Start/Stop Button
-btnLeft.addEventListener('click', function () {
-
+btnLeft.addEventListener("click", function() {
+  if (getElementsByClassName(start)) {
+    startClick();
```
ta-web-paris: This should be `chronometer.startClick()`. Remember that you are calling a method of the class Chronometer, and in line 1 we have created an instance and saved it in the variable chronometer

starter-code/javascript/main.js
```diff
@@ -1,60 +1,43 @@
...
+    Node.innerText("RESET");
+  } else if (getElementsByClassName(reset)) {
+    stopClick();
```
ta-web-paris: Same here

PR #66, opened 2018-01-15
PAR - Ngoc & Arthur JS

Ngoc0905:
Day 6 - Exercise Chronometer

### DIFF
--- a/starter-code/javascript/chronometer.js
+++ b/starter-code/javascript/chronometer.js
@@ -1,39 +1,56 @@
 // Constructor
-// function Chronometer() {
+function Chronometer() {
+    this.currentTime = 0;
+    this.intervalId = 0;
+}
 
-// }
+Chronometer.prototype.startClick = function () {
+    var that = this;
+    setInterval(function () {
+        console.log(that.currentTime);
+        that.currentTime++;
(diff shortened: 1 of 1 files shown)


starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,57 @@
...
+    if (value < 10) {
+        return "0" + value;
+    } else {
+        return "" + value;
+    }
+};
+
+Chronometer.prototype.setTime = function() {
+    var that = this;
+    this.startClick = setInterval(function(){
```
ta-web-paris: Instead of starting a new interval here, you should just call the `setTime` method in the original interval which is defined in `startClick` method

starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,57 @@
...
+function Chronometer() {
+    this.currentTime = 0;
+    this.intervalId = 0;
+}
+
+Chronometer.prototype.startClick = function () {
+    var that = this;
+    setInterval(function () {
+        console.log(that.currentTime);
+        that.currentTime++;
```
ta-web-paris: Here you should call the `setTime` method

PR #79, opened 2018-01-16
BCN - SHENG & LUIS

luisacherrera:
@therealsheng


starter-code/javascript/main.js
```diff
@@ -10,11 +10,15 @@ var milUni      = document.getElementById('milUni');
 
 
 function printTime() {
-
+    if (Chronometer.setMinutes() > 10) {
```
tawebbcn: Be careful, your variable is chronometer, not Chronometer (that is the constructor).

PR #80, opened 2018-01-16
BCN - David & Vlad

vladcio:
@davidalexteixeira


starter-code/javascript/chronometer.js
```diff
@@ -1,27 +1,45 @@
...
+
+function Chronometer(){
+    this.currentTime = 0;
+    this.intervalId = 0;
+}
+
+Chronometer.prototype.startClick = function () {
+    var self = this;
+
+    self.intervalId = window.setInterval(function() {
```

PR #81, opened 2018-01-16
BCN - Victor & Jana

jjanastinn:
@victorpellerin

### DIFF
--- a/starter-code/javascript/chronometer.js
+++ b/starter-code/javascript/chronometer.js
@@ -1,27 +1,39 @@
 // Constructor
-// function Chronometer() {
+function Chronometer() {
+    this.currentTime = 0;
+    this.intervalId = 0;
+};
 
-// }
 
-// Chronometer.prototype.startClick = function () {
+Chronometer.prototype.startClick = function () {
+    var self = this;
+    window.setInterval(function() {
+        self.currentTime = self.currentTime + 1;
+    }, 1000); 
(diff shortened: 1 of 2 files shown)


starter-code/javascript/chronometer.js
```diff
@@ -1,27 +1,39 @@
...
-// };
+function Chronometer() {
+    this.currentTime = 0;
+    this.intervalId = 0;
+};
+
+
+Chronometer.prototype.startClick = function () {
+    var self = this;
+    window.setInterval(function() {
```
tawebbcn: You don't need window here. But you need to store the interval in a variable for being able to clean it later:
```
self.intervalId = setInterval(function () {
```

starter-code/javascript/chronometer.js
```diff
@@ -1,27 +1,39 @@
...
+    }, 1000); 
+}
+
+Chronometer.prototype.setMinutes = function () {
+  return Math.floor(this.currentTime / 60);
+};
+
+Chronometer.prototype.setSeconds = function () {
+  // var minutes = Math.floor(this.currentTime / 60);
+  return this.currentTime - this.setMinutes() * 60;
```
tawebbcn: Here you transform the minutes into seconds, but what you really want is to calculate the remaining seconds. One way is:
```
return self.currentTime % 60;
```

PR #82, opened 2018-01-16
BCN - Danielle  & Alla

allarom:
@danielleleb

### DIFF
--- a/starter-code/javascript/chronometer.js
+++ b/starter-code/javascript/chronometer.js
@@ -1,26 +1,37 @@
 // Constructor
-// function Chronometer() {
+ function Chronometer() {
+    this.currentTime = 0;
+    this.intervalId = 0;
+    this.clicked = false;
+}
 
-// }
+ Chronometer.prototype.startClick = function () {
+    var self = this;
+    self.clicked = true;
+    self.intervalId = setInterval(function() {
+        self.currentTime++ ;
+    }, 1000);
+ };
 
-// Chronometer.prototype.startClick = function () {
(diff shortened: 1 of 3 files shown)


starter-code/javascript/chronometer.js
```diff
@@ -1,26 +1,37 @@
...
-  
-// };
-
-// Chronometer.prototype.twoDigitsNumber = function () {
-  
-// };
+ function Chronometer() {
+    this.currentTime = 0;
+    this.intervalId = 0;
+    this.clicked = false;
```
tawebbcn: Better get rid of that.

starter-code/javascript/chronometer.js
```diff
@@ -1,26 +1,37 @@
...
+}
+
+ Chronometer.prototype.startClick = function () {
+    var self = this;
+    self.clicked = true;
+    self.intervalId = setInterval(function() {
+        self.currentTime++ ;
```
tawebbcn: What about calling setTime after line 12?

starter-code/javascript/chronometer.js
```diff
@@ -1,26 +1,37 @@
...
+};
+
+ Chronometer.prototype.setSeconds = function () {
+    var self = this;
+    return self.currentTime % 60;
+ };
+
+Chronometer.prototype.twoDigitsNumber = function (value) {
+    var self = this;
+    if (value.indexOf) {
```
tawebbcn: What are you trying to do with this?

PR #83, opened 2018-01-16
BCN - Martin & Santi

santironhacker:
@martinX308

### DIFF
--- a/starter-code/javascript/chronometer.js
+++ b/starter-code/javascript/chronometer.js
@@ -1,39 +1,53 @@
 // Constructor
-// function Chronometer() {
+function Chronometer() {
+    this.currentTime = 0;
+    this.intervalId = 0;
+}
 
-// }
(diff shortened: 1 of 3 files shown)


starter-code/javascript/chronometer.js
```diff
@@ -1,39 +1,53 @@
...
+                                            }, 1000);
+
+};
+
+Chronometer.prototype.setMinutes = function () {
+    return Math.floor(this.currentTime/60); 
+};
+
+Chronometer.prototype.setSeconds = function () {
+    return  this.currentTime - Math.floor(this.currentTime/60)*60;
```
tawebbcn: Your solution is fine, but there are easier ways to do it. First, you are repeating your code, so you can use your function instead:
```
return  this.currentTime - this.setMinutes() * 60;
```
However, there is a better approach. With your operations you are calculating the remainder, then go for it:
```
return  this.currentTime % 60;
```

PR #84, opened 2018-01-16
BCN / Oscar y Giovani

oscarjorgejorge:
@JoniWanKenoBi

### DIFF
--- a/starter-code/javascript/chronometer.js
+++ b/starter-code/javascript/chronometer.js
@@ -1,39 +1,60 @@
 // Constructor
-// function Chronometer() {
+function Chronometer() {
+    this.currentTime;
+    this.intervalId;
+}
 
-// }
+Chronometer.prototype.startClick = function () {
+    var self = this;
+    self.intervalId = setInterval(function () {
+        self.currentTime++;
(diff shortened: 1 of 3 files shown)


starter-code/javascript/chronometer.js
```diff
@@ -1,39 +1,60 @@
...
-// };
+function Chronometer() {
+    this.currentTime;
+    this.intervalId;
+}
+
+Chronometer.prototype.startClick = function () {
+    var self = this;
+    self.intervalId = setInterval(function () {
+        self.currentTime++;
```
tawebbcn: What about calling setTime after line 10?

starter-code/javascript/chronometer.js
```diff
@@ -1,39 +1,60 @@
...
+  var self = this;
+  return self.currentTime%60;
+};
+
+Chronometer.prototype.twoDigitsNumber = function (digits) {
+  var self = this;
+if (digits.length === 1) {
+    return "0" + digits;
+    } else {
+        return digits;
```
tawebbcn: Be careful, this won't be a string if digits is a number.

PR #86, opened 2018-01-16
BCN - PERE- DANI

DTDK:
@pereamoros

### DIFF
--- a/starter-code/javascript/chronometer.js
+++ b/starter-code/javascript/chronometer.js
@@ -1,40 +1,61 @@
 // Constructor
-// function Chronometer() {
+function Chronometer() {
+  this.currentTime = 0;
+  this.intervalId = 0;
 
-// }
+ }
 
-// Chronometer.prototype.startClick = function () {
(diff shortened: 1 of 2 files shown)


starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,61 @@
...
 
-// }
+ }
 
-// Chronometer.prototype.startClick = function () {
+ Chronometer.prototype.startClick = function () {
+    var self = this;
 
-// };
+    self.intervalId = window.setInterval(function() {
```

starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,61 @@
...
-// Chronometer.prototype.twoDigitsNumber = function () {
-  
-// };
+ /*function twoDigitsNumber(value){
+      return value.padStart(2, "0");
+ }*/
 
-// Chronometer.prototype.setTime = function () {
+ Chronometer.prototype.twoDigitsNumber = function (value) {
+    return ("0" + value).slice(-2);
```
tawebbcn: Clever solution!

PR #87, opened 2018-01-16
BCN - Rodrigo & Stephie

rodrigodaoud:
@kissy89

### DIFF
--- a/starter-code/javascript/chronometer.js
+++ b/starter-code/javascript/chronometer.js
@@ -24,13 +24,12 @@ Chronometer.prototype.setSeconds = function () {
     return self.currentTime % 60;
 };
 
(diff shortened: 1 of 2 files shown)


starter-code/javascript/chronometer.js
```diff
@@ -1,23 +1,36 @@
...
-// };
+function Chronometer() {
+    this.currentTime = 0;
+    this.intervalId = 0;
+}
+
+Chronometer.prototype.startClick = function () {
+    var self = this;
+
+    self.intervalId = window.setInterval(function(){
```

starter-code/javascript/chronometer.js
```diff
@@ -1,23 +1,36 @@
...
+        return self.currentTime += 1;
```
tawebbcn: What about calling setTime after line 11?

PR #89, opened 2018-01-16
BCN JAN 18 - MATT & PEDRO!!!! We done this clock thing innit...nearly

mattmegarry:
@pedrodiogoneto

### DIFF
--- a/starter-code/javascript/chronometer.js
+++ b/starter-code/javascript/chronometer.js
@@ -1,27 +1,60 @@
 // Constructor
-// function Chronometer() {
+function Chronometer() {
+        
+    this.currentTime = 0;
+    this.intervalId = 0;
 
-// }
+}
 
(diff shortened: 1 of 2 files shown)


starter-code/javascript/chronometer.js
```diff
@@ -1,27 +1,60 @@
...
-// }
+}
 
-// Chronometer.prototype.startClick = function () {
 
-// };
+Chronometer.prototype.startClick = function () {
+    var self = this;
+    //Reminder: self.intervalId = (if OK not, how come?)
+    window.setInterval (function () {
```

starter-code/javascript/chronometer.js
```diff
@@ -1,27 +1,60 @@
...
+        self.setTime();
```
tawebbcn: Better call this function after increasing currentTime.

PR #125, opened 2018-03-26
BCN - Jordi

jordicalzada:


starter-code/javascript/chronometer.js
```diff
@@ -1,23 +1,33 @@
...
+function Chronometer() {
+    this.currentTime = 0;
+    this.intervalId = null;
 
-// }
+}
 
-// Chronometer.prototype.startClick = function () {
+Chronometer.prototype.startClick = function () {
+    this.intervalId = window.setInterval(this.startClick,1000);
```
tawebbcn: What you should put on the first setInterval argument is the callback that you want to run every `1000`ms.
So you should have something like:
```
this.intervalId = window.setInterval(function () {
    this.currentTime++
  },1000);
```

Right now what you are doing is calling the method `startClick` every 1000ms. That means that you are creatin an interval each 1000ms since that is what you have in your line 9

starter-code/javascript/chronometer.js
```diff
@@ -1,23 +1,33 @@
...
 
-// };
+};
 
-// Chronometer.prototype.setMinutes = function () {
-  
-// };
+Chronometer.prototype.setMinutes = function () {
+
+    return this.currentTime%60;
```
tawebbcn: For the minutes you can do `Math.floor(this.currentTime / 60)`

starter-code/javascript/chronometer.js
```diff
@@ -1,23 +1,33 @@
...
+
+};
+
+Chronometer.prototype.setSeconds = function () {
 
-// Chronometer.prototype.setSeconds = function () {
+    return this.currentTime;
```
tawebbcn: For the second you can do `Math.floor( this.currentTime % 60 )`

starter-code/javascript/chronometer.js
```diff
@@ -27,9 +37,11 @@
 
 // };
 
-// Chronometer.prototype.stopClick = function () {
+Chronometer.prototype.stopClick = function () {
+
+    this.intervalId = window.clearInterval(this.currentTime);
```
tawebbcn: For clear an interval you have to pus in the argument the id of the interval. And you don't need to save the return value. So you should have something like:
```
window.clearInterval(this.intervald);
```

PR #133, opened 2018-03-27
BCN- RAPHA MONTENEGRO

raphamontenegro:
Had some issues wit hth e double digit.

### DIFF
--- a/starter-code/javascript/chronometer.js
+++ b/starter-code/javascript/chronometer.js
@@ -1,40 +1,71 @@
 // Constructor
-// function Chronometer() {
 
-// }
+function Chronometer() {
+  this.intervalId = null;
+  this.currentTime = 0;
+}
 
-// Chronometer.prototype.startClick = function () {
+Chronometer.prototype.startClick = function () {
+  var that = this;
+  this.intervalId = setInterval(function(){
+    that.currentTime += 1;
+  }, 1000)
+};
 
-// };
+Chronometer.prototype.setMinutes = function () {
+    this.minutes = this.currentTime / 60;
+    return Math.floor(this.minutes);
+};
 
-// Chronometer.prototype.setMinutes = function () {
-  
-// };
+Chronometer.prototype.setSeconds = function () {
+  if (this.currentTime > 60) {
+    this.seconds = this.currentTime % 60;
+  } else {
+    this.seconds = this.currentTime;
+  }
+  return this.seconds;
+};
 
-// Chronometer.prototype.setSeconds = function () {
-  
-// };
+Chronometer.prototype.twoDigitsNumber = function () {
+  /*if (this.minute < 10) {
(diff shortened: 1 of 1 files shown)


starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,71 @@
...
-// };
+Chronometer.prototype.setMilliseconds = function () {
+  this.milliseconds = this.currentTime * 100;
+  return this.milliseconds;
+};
 
-// Chronometer.prototype.setMilliseconds = function () {
+Chronometer.prototype.stopClick = function () {
+ var that = this;
+  clearInterval(that.intervalId)
```

tawebbcn: With ```that``` it's ok, but we are using ```self``` as a convention.

starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,71 @@
...
 
-// Chronometer.prototype.setMinutes = function () {
-  
-// };
+Chronometer.prototype.setSeconds = function () {
+  if (this.currentTime > 60) {
+    this.seconds = this.currentTime % 60;
+  } else {
+    this.seconds = this.currentTime;
+  }
```
tawebbcn: You just need to ``` return this.currentTime % 60```

tawebbcn: eg. 3 % 60 = 3

starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,71 @@
...
+    this.minutes = this.currentTime / 60;
+    return Math.floor(this.minutes);
+};
 
-// Chronometer.prototype.setMinutes = function () {
-  
-// };
+Chronometer.prototype.setSeconds = function () {
+  if (this.currentTime > 60) {
+    this.seconds = this.currentTime % 60;
```
tawebbcn: this.seconds is not declared

starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,71 @@
...
-// function Chronometer() {
 
-// }
+function Chronometer() {
+  this.intervalId = null;
+  this.currentTime = 0;
+}
 
-// Chronometer.prototype.startClick = function () {
+Chronometer.prototype.startClick = function () {
```
tawebbcn: Niiice

starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,71 @@
...
 
-// Chronometer.prototype.stopClick = function () {
-  
-// };
+Chronometer.prototype.splitClick = function () {
+//Too complicated for me right now. I'd love to see it on the exercise review
+};
 
-// Chronometer.prototype.resetClick = function () {
+function main() {
```
tawebbcn: This function isn't doing anything (and you don't need it)

PR #137, opened 2018-03-27
BCN Fabian 

fscheugenpflug:
I had problems with Jasmine. Additionally I am not to confident about JS-HTML connection yet.

### DIFF
--- a/starter-code/javascript/chronometer.js
+++ b/starter-code/javascript/chronometer.js
@@ -1,40 +1,53 @@
 // Constructor
-// function Chronometer() {
+function Chronometer() {
+  this.currentTime = 0;
+  this.intervalId = 0;
+}
 
-// }
+Chronometer.prototype.startClick = function () {
+  var self = this;
+  var callbackFunction = function (){
+    self.currentTime++ 
+  };
+  intervalID = window.setInterval(callbackFunction, 1000);
+};
 
-// Chronometer.prototype.startClick = function () {
+Chronometer.prototype.setMinutes = function () {
+  return Math.floor(this.currentTime/60);
+};
 
-// };
+Chronometer.prototype.setSeconds = function () {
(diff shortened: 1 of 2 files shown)


starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,53 @@
...
+  this.currentTime = 0;
+  this.intervalId = 0;
+}
+
+Chronometer.prototype.startClick = function () {
+  var self = this;
+  var callbackFunction = function (){
+    self.currentTime++ 
+  };
+  intervalID = window.setInterval(callbackFunction, 1000);
```

tawebbcn: Then in the main.js you have to call this inside 
```
btnLeft.addEventListener('click', function () {
   chronometer.startClick();
});
```

starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,53 @@
...
-
-// };
+function Chronometer() {
+  this.currentTime = 0;
+  this.intervalId = 0;
+}
+
+Chronometer.prototype.startClick = function () {
+  var self = this;
+  var callbackFunction = function (){
```
tawebbcn: next step: declaring the callback inside the setInterval. It will be something like:
```
     self.currentTime += 1;
     self.setTime();
}, 1000);
```

starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,53 @@
...
+Chronometer.prototype.setTime = function () {
+    
+};
+
+Chronometer.prototype.setMilliseconds = function () {
+
+};
+
+Chronometer.prototype.stopClick = function () {
+  clearInterval(this.intervalId);
```
tawebbcn: :heavy_check_mark:

PR #138, opened 2018-03-27
BCN Dauv

dauvbafna:


starter-code/javascript/chronometer.js
```diff
@@ -1,39 +1,53 @@
...
+}
+
+Chronometer.prototype.startClick = function(){
+  var self = this;
+
+  function increment(){
+    self.currentTime += 1;
+    self.setTime();
+  }
+  this.intervalId = setInterval(increment, 1000);
```
tawebbcn: Niice!

tawebbcn: Then in the main.js you have to call this inside 
```
btnLeft.addEventListener('click', function () {
   chronometer.startClick();
});
```

starter-code/javascript/chronometer.js
```diff
@@ -1,39 +1,53 @@
...
+Chronometer.prototype.twoDigitsNumber = function (time){
+  if(time >= 0 && time<10){
+    time = "0"+time;
+    return time;
+  } else { 
+    return time+ "";
+  }
+}
+
+ Chronometer.prototype.setTime = function () {
```

starter-code/javascript/chronometer.js
```diff
@@ -1,39 +1,53 @@
...
-// };
+  function Chronometer() {
+    this.currentTime = 0;
+    this.intervalId =0;
+}
+
+Chronometer.prototype.startClick = function(){
+  var self = this;
+
+  function increment(){
```
tawebbcn: next step: declaring the callback inside the setInterval. It will be something like:
```
     self.currentTime += 1;
     self.setTime();
}, 1000);
```

PR #156, opened 2018-05-28
BER Silvio and Laure

LaureDas:


starter-code/javascript/main.js
```diff
@@ -8,24 +8,41 @@ var secUni      = document.getElementById('secUni');
 var milDec      = document.getElementById('milDec');
 var milUni      = document.getElementById('milUni');
 
+setInterval(function(){
```
ta-web-ber: you don't need this setInterval. You can use the setInterval inside chronometer.js. To do that, inde setTime method:
```
Chronometer.prototype.setTime = function () {
  var minutes = this.twoDigitsNumber(this.setMinutes());
  var seconds = this.twoDigitsNumber(this.setSeconds(minutes));
  printTime(minutes, seconds);
}
```

PR #157, opened 2018-05-28
BER-Hannes-Reinsch

hannesreinsch:


starter-code/javascript/chronometer.js
```diff
@@ -1,23 +1,42 @@
...
+Chronometer.prototype.setMinutes = function() {
+    return Math.floor(this.currentTime / 60);
+  };
 
-// Chronometer.prototype.twoDigitsNumber = function () {
-  
-// };
+
+
+ Chronometer.prototype.setSeconds = function () {
```
ta-web-ber: Here you can pass minutes as a parameter:
```
Chronometer.prototype.setSeconds = function (minutes) {
  return this.currentTime -minutes*60);
 };
```
or 
```
Chronometer.prototype.setSeconds = function () {
  return this.currentTime % 60;
 };

PR #160, opened 2018-05-28
BER - John Fish - Chrono project - done

j31:
Seems to work well with all the spec-ed features.  When I added milliseconds, the Jasmine test is now producing 2 of the same errors, "can't read null type of innerText", but I don't see where that could be happening... If someone looks at this, I would like to understand how to track down this kind of error from Jasmine and find it in the code.


ta-web-ber:
@mc100s

PR #233, opened 2018-08-06
BER-Tim-Ana

tlebon:


ta-web-ber:
Good for the first iteration, but then you should use create an instance of the `Chronometer` class and use its class methods to manipulate the DOM element and display minutes and seconds on the watch.

PR #241, opened 2018-08-06
BER - Andrei - Done

AndreiFitai:


ta-web-ber:
Excellent job, Andrei!

The only thing to change is al line 16 of `chronometer.js`:
```javascript
  // return parseInt( ((this.currentTime) / (100*60))%60 );
  return parseInt( (this.currentTime) / (100*60)  );
```
if you keep your code you're not displaying the minutes. 

Silvio

PR #281, opened 2018-09-23
[WEB-PT-MAD2] AnaEstrada

aemartos:
I've done the exercise but the milliseconds. I'm freaking out with that. I'm doing the bootstrap exercise and then I'll return with that lovely thing!  😅🔫

EDIT: Done. I was doing the calculation wrong. I've had problems with the change between time measures. Btw when I implement the milliseconds, Jasmine doesn't valid the exercise.

### DIFF
--- a/starter-code/javascript/chronometer.js
+++ b/starter-code/javascript/chronometer.js
@@ -63,6 +63,10 @@ Chronometer.prototype.resetClick = function () {
   this.currentTime = 0;
 };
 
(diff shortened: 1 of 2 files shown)


ta-web-mad:
Good job Ana!! 😊
Don't worry about Jasmine, it expects the currentTime to be in seconds  , so as you are counting each 10 miliseconds it does't approve , but your code works !

#checked #closed Niko 💪🏻.

PR #293, opened 2018-10-06
work js

sylvia-moreno:


starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,68 @@
...
+};
 
-// Chronometer.prototype.setTime = function () {
 
-// };
+Chronometer.prototype.twoDigitsNumber = function(number) {
+    if (number < 10) {
+        return number = '0' + number;
+    } else {
+        return number += '';
```
ta-webpt-par: Note that since you return in your `if`, the useless keyword is useless. To make your code cleaner you could have done this:
```javascript
    if (number < 10) {
        return number = '0' + number;
    } 
    return number.toString();
```

ta-webpt-par:
Great work !

PR #298, opened 2018-10-14
[PAR-PT] JustineZahiri

justinezahiri:
iteration 1 & 2 completed but displaying the time in a "hacky" way (by calling in chronometer.js printTime function declared in main.js)
@ta-webpt-par


ta-webpt-par:
To avoid your `printTime() function` in the `Chronometer class` you should have used another `setInterval` in the `main.js` file.

PR #310, opened 2018-10-22
[SAO FT1 - PP - W2D1] Jeff Smith & Carol Park

jmstranslate:
Extremely difficult.


thebinaryfelix:
Nice implementation of ES6 class and methods, but still missing HTML - JS integration (showing the numbers on the counter)

PR #311, opened 2018-10-22
[SAO FT1 - PP - W2D1] Mohsan Abbas & Ismael Paiva

mohsanabbas:


thebinaryfelix:
The HTML/CSS implementation is missing, but the JS logic looks good.
Note: start using eslint to standardise the code. At least before submitting, format the code in VS Code.

PR #330, opened 2018-10-29
AMS jesco wuester

jescowuester:


IronhackAMS:
Ha Jesco!,
First one of the group that I come across that uses a constructor function and prototypical inheritance, nice! For you then as a bonus if you like, try to solve it with the use of method abstraction on objects?
Kido

PR #331, opened 2018-10-29
AMS- Lyudmyla Baron

lxs07a:


starter-code/javascript/main.js
```diff
@@ -8,53 +8,67 @@ var secUni      = document.getElementById('secUni');
...
-
+  if (btnRight.classList.contains("reset")) {
+    chronometer.reset();
+  } else {
+    split();
+  }
 });
+
+
+function update (minutes, seconds, miliseconds) {
```
IronhackAMS: the parameters placeholders here are not invoked so you can leave it out

IronhackAMS:
Nice work Lyudmyla using abstraction with methods and DOM manipulation!
There are generally 3 ways to go about this problem: 1) using abstraction with methods. 2) using constructor functions and prototypical inheritance or 3) use a class constructor. 
You did nr 1) ;)
If you like and as a bonus, try to solve this assignment using constructor functions or a class constructor:)
Kido

PR #336, opened 2018-10-29
AMS Dea Kristina

Deakristina:


IronhackAMS:
Very well done! Your code is really readable. If you'd like to challenge yourself a little try to refactor your chorometer and make use a class constructor.

PR #342, opened 2018-10-29
AMS Matt Hamers 

mattviy:
finalized chronometer


IronhackAMS:
Very good job! Note that you're not reusing the code on line 6,7 and 8 of your main.js. You can either use the properties, for example at line 15, 16 and 17 respectively, or get rid of it.

PR #351, opened 2018-10-30
AMS Ana Kelly

anakellyc:


IronhackAMS:
Ola Ana!
Nice work using method abstraction on objects! If you like and as a bonus, try to solve this assignment using constructor functions and prototypical inheritance or by use of a class constructor.
Kido

PR #447, opened 2019-02-09
[AMS] - Fedde

Fedduh:
Adding the miliseconds messed up the Jasmin validations :)

### DIFF
--- a/starter-code/javascript/chronometer.js
+++ b/starter-code/javascript/chronometer.js
@@ -6,6 +6,7 @@ function Chronometer() {
 
 Chronometer.prototype.startClick = function () {
   // because of scoping, make a copy of this
+  // console.log("startClick called");
   var self = this;
   self.intervalId = setInterval(function() {
     self.currentTime ++;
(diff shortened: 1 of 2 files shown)


starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,78 @@
...
-  
-// };
+Chronometer.prototype.setTime = function () {
+  // Taking the current time and set it in minutes and seconds (2 digits)
+  var minutes = this.twoDigitsNumber(this.setMinutes());
+  var seconds = this.twoDigitsNumber(this.setSeconds());
+  var milisec = this.twoDigitsNumber(this.setMilliseconds());
+  // display seconds
+  secDec.innerHTML = seconds.substr(0, 1);
+  secUni.innerHTML = seconds.substr(1, 1);
```
lienekechee: `seconds.charAt(0)` and `seconds.charAt(1)` are preferable here since you just want to get one character, and `substr()` or rather `.substring()` should be used to find sections of string. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr), `.substr()`, is anyway being deprecated.

starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,78 @@
...
-  
-// };
+Chronometer.prototype.twoDigitsNumber = function (number) {
+  // if one digit, add a leading 0
+  if (number < 10) {
+    return "0" + number.toString();
+  } else {
+    return number.toString();
+  }
+};
```
lienekechee: Since you seem to have an appreciation for concise code you might be interested in using [ternary operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) for simple conditional statements: 
```
return (number < 10) ? "0" + number.toString() :  number.toString()
```

PR #456, opened 2019-02-11
[AMS] - Isabela

isabeladowsley:


starter-code/javascript/main.js
```diff
@@ -7,54 +7,86 @@ var secDec      = document.getElementById('secDec');
...
+    console.log(btnLeft.innerHTML);
+    if (btnLeft.innerHTML === "START
") {    
+        chronometer.startClick();
+        printTime();
+        setStopBtn();
+    } else {    
+        chronometer.stopClick();
+        setStartBtn();
+    }
```
isabeladowsley: For me, it seems that the counter itself it's working great, but for some reason, it's not printing. I've tried to check through different ways but couldn't find where does the mistake is.

starter-code/javascript/main.js
```diff
@@ -7,54 +7,86 @@ var secDec      = document.getElementById('secDec');
...
 
+btnRight.addEventListener('click', function () {
+    if (btnRight.innerHTML == "RESET") {
+        chronometer.resetClick();
+        setSplitBtn();
+    } else {
+        printSplit();
+        setResetBtn();
+    }  
 });
```
isabeladowsley: Also the splits are not been printed in a proper way.

PR #457, opened 2019-02-11
Amsterdam - Guido Carucci 

guido-an:


starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,59 @@
...
-// };
-
-// Chronometer.prototype.stopClick = function () {
-  
-// };
 
-// Chronometer.prototype.resetClick = function () {
 
+ // Chronometer.prototype.setMilliseconds = function () {
+
```
lienekechee: Please remove unused functions, whitespace and comments :)

starter-code/javascript/main.js
```diff
@@ -7,54 +8,78 @@ var secDec      = document.getElementById('secDec');
...
+     btnRight.style.backgroundColor = "#0851ab"; 
+     btnLeft.style.backgroundColor = "#f14949"; 
+
+ } else  {
+     chrono.stopClick();           // change to "START" and "RESET" 
+     btnLeft.innerHTML = "START";
+     btnLeft.style.backgroundColor = "#5fca5f"; 
+     btnRight.innerHTML = "RESET";  
+     btnRight.style.backgroundColor = "#908e8e"; 
+ }
```
lienekechee: If you look into the styles file, you will see that the classes `.btn.start` and`.btn.stop` contain the background colors for the buttons.  So rather than setting the background color here, its better to toggle the classes.

starter-code/javascript/main.js
```diff
@@ -7,54 +8,78 @@ var secDec      = document.getElementById('secDec');
 var secUni      = document.getElementById('secUni');
 var milDec      = document.getElementById('milDec');
 var milUni      = document.getElementById('milUni');
+var splits      = document.getElementById('splits');
 
+function showTime() {   // display time on the chronometer
+    setInterval(printTime, 100)
+}
```
lienekechee: So I know we looked at this together, but I must say while is one solution, it is really not the most efficient way for doing this, because this `setInterval` here is being cleared anywhere and also it means you have two concurrent processes running.  You have access to `document` in your `chronometer.js` file as well.  Or you could pass the `printTime` function to your chronometer `startClick` function:
```
if (btnLeft.innerHTML == "START") {   // buttons on load: START and RESET
    chrono.startClick(printTime);
```
then in chronometer.js:
```
 Chronometer.prototype.startClick = function (printTime) {  // Start Click
    var that = this
        that.currentTime += 1;
        console.log(that.currentTime)
        that.setTime();
        printTime()
     }, 1000)

};
```

PR #475, opened 2019-02-11
completed with bonuse

debasrid:


starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,64 @@
...
+  this.intervalId = setInterval(function () {   // Every 10 milisec counter increment by 1
+    that.currentTime += 1;
+    console.log(that.currentTime)
+  }, 10);
+};
 
-// };
+Chronometer.prototype.setMinutes = function () {
+  return this.minutes = Math.floor(Math.floor(this.currentTime / 100) / 60);  //calculated minutes from milisec
+};
```
lienekechee: Don't think we need to use `Math.floor` multiple times here.  Just the outer ones should be enough.  Also, have a look into the `%` operator.  This operator return the remainder after dividing by a given number.  So to calculate the times, I think this is more efficient:
```
this.seconds = (this.currentTime % this.minutes) / 100 
// Get remainder of milliseconds after minutes are calculated;
// Divide by 100 to calculate seconds
```

starter-code/javascript/chronometer.js
```diff
@@ -1,40 +1,64 @@
...
-  
-// };
+Chronometer.prototype.resetClick = function () {
+  this.currentTime = 0;
+};
+
+// ------------------------------------------------------------
 
-// Chronometer.prototype.resetClick = function () {
+// Chronometer.prototype.setTime = function () {
```
lienekechee: Please always remove unused code, comments and whitespace before committing!

starter-code/javascript/main.js
```diff
@@ -1,60 +1,122 @@
...
+var minDec = document.getElementById('minDec');
+var minUni = document.getElementById('minUni');
+var secDec = document.getElementById('secDec');
+var secUni = document.getElementById('secUni');
+var milDec = document.getElementById('milDec');
+var milUni = document.getElementById('milUni');
+
+function showTime() {
+    setInterval(printTime, 10)  // to call printTime function every 10 milisec
+}
```
lienekechee: So you and Guido have the same solution, and although it is a solution, it is not ideal because now we can two `setInterval`s running concurrently.  You have two better options: first you could set the html value directly from the Chronometer.js file, but then your code would not be properly modularized.  Secondly, you could pass the `printTime` function as an argument to the `chrono.startClick` function.  So something like:
```
Chronometer.prototype.startClick = function(printTime) {
  this.intervalId = setInterval((function() {
    this.currentTime++
    this.setTime()
    printTime()
  }).bind(this), 1000);
};
```
This way, your data is defined and structured in one file, and your executable javascript is in another.

starter-code/javascript/main.js
```diff
@@ -1,60 +1,122 @@
...
+        if (btnRight.innerHTML == "RESET") {
+        btnLeft.innerHTML = "START"
+        btnLeft.style.backgroundColor = "#5fca5f"
+        chrono.currentTime = 0;
+        chrono.stopClick();                 // setInterval stopped
+        clearSplits();                      // splits text cleared
+
+    }
+    else {
+
```
lienekechee: Please avoid leaving unnecessary white space in your code!  Best practice and good habits regarding style and consistency are really important things to develop early on, it will help you in your learning, I promise :)

starter-code/javascript/main.js
```diff
@@ -1,60 +1,122 @@
...
 
 }
 
-function setStopBtn() {
+function setStopBtn() {                     //when clicked button color & text chnged
+    btnLeft.innerHTML = "START"
+    btnLeft.style.backgroundColor = "#5fca5f"
+    btnRight.innerHTML = "RESET"
+    btnRight.style.backgroundColor = "#908e8e"
+
```
lienekechee: The styles for the buttons are defined by the class names `.btn.left` and `.btn.right`.  So what you do here is better achieved by simply adding an removing these classes from your button elements ;)

PR #609, opened 2019-08-01
[Amsterdam] Augusta Marques

lotofcaffeine:


starter-code/javascript/chronometer.js
```diff
@@ -1,10 +1,79 @@
...
-  // stopClick() {}
-  // resetClick() {}
-  // splitClick() {}
+
+    constructor() {
+        this.currentTime = 0;
+        this.Milisecond = 0;
+        this.miliIntervalId = 0;
+        this.intervalId = 0;
+        this.status = "stopped";
```
suwigyarathore: No need of this extra variable. You can track if the watch is running or not just by checking if interval id is null or not after u clearinterval.

starter-code/javascript/chronometer.js
```diff
@@ -1,10 +1,79 @@
...
+    }
+    start() {
```
suwigyarathore: same goes for it

starter-code/javascript/chronometer.js
```diff
@@ -1,10 +1,79 @@
...
+    }
+    stop() {
+
+        this.status = "stopped";
+
+    }
+    startClick(callback) {
+
+        this.start();
+        this.miliIntervalId = setInterval(() => {
```
suwigyarathore: try using only one interval and do maths to get different units out of it

suwigyarathore: This function will very soon make ur browser freeze ... u actually need a supercomputer to execute one function 1000 times in 1 sec :)

PR #615, opened 2019-08-13
Chronometer exercise

studiobicker:


starter-code/javascript/chronometer.js
```diff
@@ -1,10 +1,47 @@
...
-  // resetClick() {}
+  constructor(show) {
+    this.currentTime = 0;
+    this.intervalId = null;
+    this.showTime = show
+  }
+  
+  startClick() {
+    this.intervalId = setInterval(()=>{
+      //debugger;
```
suwigyarathore: you can remove these comments

starter-code/javascript/chronometer.js
```diff
@@ -1,10 +1,47 @@
 class Chronometer {
```
suwigyarathore: It would be nice to see milliseconds also

PR #620, opened 2019-08-15
PAR-FTWD-Justine & Niraja

niraja426:


ta-web-paris:
Nice! (you would just need to set currentTime at 0 to clear the 3 jasmine errors, but that's a detail)

PR #676, opened 2019-10-24
Complete

venkat-raghulan:


ta-web-paris:
Good job for the bonus! Though I think you might have broken something when adding the milliseconds, Jasmine has 7 errors - when you get the time try to fix your code :)

venkat-raghulan:
Guess Jasmine says errors because I am using setinterval with an increment
of 10 instead of 1000 to get the milliseconds.
Is there any other way to do it?

Thanks

PR #677, opened 2019-10-24
[PAR WDFT] - Marion & Amine 

aminelahrichi:


ta-web-paris:
Very nice job! :) Try to clear the last two jasmine tests if you get the time it should be easy

PR #678, opened 2019-10-24
[PAR WDFT] Evgeny

ESR93:


ta-web-paris:
Chronometer looks good :) Check if you can make Jasmine all green if you have time!

PR #715, opened 2019-11-23
[ING-FT-112019] - Tomasz

twalasek:
Note: 
In the bonus part, the IronHack instructions say about milliseconds. But, the HTML clock contains only two span tags for the "millisecond": milDec and milUni. Therefore, in my design I chose my clock to tick every 10 milliseconds (i.e. 1/100 of a second). That's OK, because ticking every millisecond would be pointless - even in athletics the smalles unit of time is 1/100 second. For eample, the current 100 m men's world record is 9.58 second.

### DIFF
--- a/starter-code/javascript/chronometer.js
+++ b/starter-code/javascript/chronometer.js
@@ -1,10 +1,32 @@
 class Chronometer {
-  constructor() {}
-  // startClick() {}
-  // getMinutes() {}
-  // getSeconds() {}
-  // twoDigitsNumber() {}
-  // stopClick() {}
-  // resetClick() {}
-  // splitClick() {}
+  constructor() {
+    this.currentTime = 0;
+    this.intervalId = 0;
+  }
+  startClick() {
+    this.intervalId = setInterval(function () 
+      {chronometer.currentTime += 1}, 10);
+  }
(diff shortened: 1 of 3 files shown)


starter-code/javascript/chronometer.js
```diff
@@ -1,10 +1,32 @@
...
-  // splitClick() {}
-}
\ No newline at end of file
+  constructor() {
+    this.currentTime = 0;
+    this.intervalId = 0;
+  }
+  startClick() {
+    this.intervalId = setInterval(function () 
+      {chronometer.currentTime += 1}, 10);
```
MarcoSantonastasi: ERROR: hard coded parameter is not allowed

starter-code/javascript/main.js
```diff
@@ -7,54 +21,100 @@ var secDec      = document.getElementById('secDec');
...
+    printMinutes();
+    printSeconds();
+    printMilliseconds();
+  }, 10);
 }
 
 function printMinutes() {
-
+  let twoDigits = chronometer.getMinutes();
+  minDec.innerHTML = twoDigits.charAt(0);
```
MarcoSantonastasi: Prefer index notaion

PR #719, opened 2019-11-25
Ironhack - AMS - WebDev - 201911 - Marc de Hoogh (to be continued)

mdehoogh:
functional without checking for test conformance


mdehoogh:
marcschronometer.js now contains the Chronometer version of yesterday that didn't pass the test. The new chronometer.js now does. It can accept (in its interval property) the number of milliseconds between successive 'ticks', and keeps track of the number of ticks in this.ticks.

PR #761, opened 2020-01-15
[PAR-WDFT] Louis & Tayssir - ongoing

louis-md:


Jepassaispar:
Good move setting the isRunning function inside the chronometer ! It's a great use of a class functionality !

PR #763, opened 2020-01-15
[PAR-WDFT] - Anthony B. - ongoing

abrebion:
Done


Jepassaispar:
It looks very good, I would just have put the `chronometer.resetClick()` function inside the btnRight event listener so the time is reset when you click on the button reset and not whenever you start the chronometer.

PR #771, opened 2020-01-16
[PARWD][012020] TAYSSIR .  [ongoing]

taytay81:


starter-code/javascript/chronometer.js
```diff
@@ -8,19 +8,19 @@ class Chronometer {
   startClick(callback) {
     this.isRunning = true;
     this.intervalId = setInterval(() => {
-      this.currentTime += 1 ;
+      this.currentTime += 1;
       console.log(this.currentTime);
       callback();
-      }, 1000);
-    }
+    }, 1000);
```
frankmarmier: Adding the callback here breaks the jasmine test because Jasmine doesn't send a callback function when she tests the startClick function, in order to fix it, you could just test if the callback is defined or not in order to execute the function.

 ```
 startClick(callback) {
    this.isRunning = true;
      this.currentTime += 1;
      console.log(this.currentTime);
      if (callback) {
        callback();
      }
    }, 1000);
  }
```

File: starter-code/javascript/chronometer.js [hunk 2]
```diff
@@ -8,19 +8,19 @@ class Chronometer {
...
 
   getMinutes() {
     let minutes = Math.trunc(this.currentTime / 60);
     return this.twoDigitsNumber(minutes);
   }
 
   getSeconds() {
-    let seconds = this.currentTime % 60 ;
+    let seconds = this.currentTime % 60;
     return this.twoDigitsNumber(seconds);
```
frankmarmier: this breaks the jasmine test aswell, you getSeconds should return a number, even though it is convenient for the lab to return a string,  the Chronometer class might be used in a different context than the DOM so it would still be good to return a number.

frankmarmier:
Nice use of the cb function for to update the DOM without using an additional setInterval, 
Apart from the change of implementation for DOM convenience, everything works perfectly.
Well done

PR #773, opened 2020-01-16
[PAR - WDFT] Remy Pham/Kathleen Domingo - ongoing

kurosaek:


Jepassaispar:
All good :) The only thing is that you had a delay in your setInterval causing your chronometer to start after 1s, and your split to be one second in advance, this is due to your callback inside the setInterval that is before the currentTime incrementation. If you put it after it will fix it :)
Great job 👍

PR #779, opened 2020-01-18
[PAR WD FT]Jonathan Oreja .  [ongoing]

Jarchiii:
I don't understand why the split button return undefined.....
the reset work

### DIFF
--- a/starter-code/javascript/main.js
+++ b/starter-code/javascript/main.js
@@ -7,25 +7,35 @@ var secDec      = document.getElementById('secDec');
 var secUni      = document.getElementById('secUni');
 var milDec      = document.getElementById('milDec');
 var milUni      = document.getElementById('milUni');
-
-
-function printTime() {
-
-}
+var splitList = document.getElementById('splits');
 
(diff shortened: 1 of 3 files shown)


starter-code/javascript/main.js
```diff
@@ -7,25 +7,35 @@ var secDec      = document.getElementById('secDec');
...
 }
 
 function printSeconds() {
+  secUni.innerHTML = chronometer.twoDigitsNumber(chronometer.getSeconds())[1];
+  secDec.innerHTML = chronometer.twoDigitsNumber(chronometer.getSeconds())[0];
+}
 
+function printTime() {
+  console.log(printMinutes()+" : "+printSeconds())
+  return printMinutes()+" : "+printSeconds()
```
ta-web-paris: The reason why you're getting undefined here is because  your functions printMinutes and printSeconds aren't returning anything, they are just changing the values of the the DOM elements, so by default they are returning undefined. 
So at the end you're returning undefined + undefined.

PR #781, opened 2020-01-18
[PAR-FT-012020] Wendy Chronometer . [ongoing]

Wendev13:


starter-code/javascript/main.js
```diff
@@ -1,60 +1,99 @@
...
+    btnLeft.className = "btn start";
 }
 
 function setResetBtn() {
-
+    btnRight.innerHTML = "RESET";
+    btnRight.className = "btn reset";
 }
 
+window.setInterval(printTime, 1000);
```
ta-web-paris: You should make the interval a little faster than the chronometer otherwise the time is not updated according to the chronometers current time value, but a second after the startClick for example.

`window.setInterval(printTime, 100);`

ta-web-paris:
Well done for the lab !

PR #845, opened 2020-04-01
PAR WDFT Sam L ongoing

samlfair:


TatijanaRajcic:
Hey Sam! Thanks for your lab! The result looks liiite !
Small things:
- if you want to avoid some errors (even if there are not blocking here), you could write an 'if' condition in your startClick method, so that, when we do not pass a callback function to the startClick method, it would still work: instead of writing "cb()", you can write "if (cb) cb()" meaning "executes the callback if there is one"
- Whenever I stop the chronometer by clicking on "reset", it resets the DOM by replacing every number by 0, but it does not reset the chronometer, since it starting again from this final time when I click on "start" again. Just use your "resetClick" method :)
```

// Reset/Split Button
  if(btnRight.classList.contains("split")) {
    printSplit();
  } else {
    chronometer.resetClick()
    minDec.innerHTML = "0"
    minUni.innerHTML = "0"
    secDec.innerHTML = "0"
    secUni.innerHTML = "0"
    milDec.innerHTML = "0"
    milUni.innerHTML = "0"
    splits.innerHTML = ""
  }
});
```

Otherwise excellent work !

PR #853, opened 2020-04-02
PAR WD FT March - Vladimir Leloup

Vlp90:
Chrono doesnt work, don't know why.


ta-web-paris:
Hey Vladimir !

Nice job overall, except for the startClick : 
Screenshot 2020-04-09 at 14 15 40

You should write : 
>> if(callback){
callback( )
}

because in the if( ) it's a parameter, putting the parentheses after the name would call the function immediately and we don't want that. If there is a callback, then we execute it.

Otherwise good !

Nina

PR #860, opened 2020-04-05
PAR WDFT 0320 Teddy ROGERS 

TeddyRod:


ta-web-paris:
Nice job Teddy :)

We'll send a solution so you guys can read it to check your code, since we will enter module 3 and it's project week we can't go over it again

Nina

PR #861, opened 2020-04-06
Louis Alibert - WDFT 0320 PAR - almost done

lalib19:
Not sure why the time is not running but the buttons still seem to work


ta-web-paris:
Hey Louis ! 

Everything js-related is working, good job ! you can still check the solutions in the cohort repository to see the code and understand it ! If anything is not clear hit me up :)

Nina

PR #922, opened 2020-06-19
 [PAR-FT-0620] Menhaj safi

Menhaj01:
Hi Tatijana,
 Many thanks for your feedback!


TatijanaRajcic:
Hey Menhaj!
Be careful, in your stopClick() method, what you should clear is the intervalId, which is the id of your setInterval, and not the currentTime which is not related at all to clearInterval.
- When you press start, the btnLeft turns to stop which is perfect, but on the right side, the button should change from "reset" to "split", and whenever you stop your chrono, the button on the right should turn to "reset" again. In order words, your event listeners should be like this;

```
// Start/Stop Button
btnLeft.addEventListener("click", () => {
  if (btnLeft.classList.contains("start")) {
    // add the stop class: classList.add
    btnLeft.classList.add("stop");
    // remove the start class : classList.remove
    btnLeft.classList.remove("start");
    // change the text into "STOP" : innerHTML
    btnLeft.innerHTML = "STOP";
    // add the split class:
    btnRight.classList.add("split");
    // remove the reset class:
    btnRight.classList.remove("reset");
    // change the text into "split"
    btnRight.innerHTML = "SPLIT";
    chronometer.startClick(printTime);
  } else {
    // add the start class: classList.add
    btnLeft.classList.add("start");
    // remove the stop class : classList.remove
    btnLeft.classList.remove("stop");
    // change the text into "START" : innerHTML
    btnLeft.innerHTML = "START";
    // add the reset class
    btnRight.classList.add("reset");
    // remove the split
    btnRight.classList.remove("split");
    // change the text into "RESET"
    btnRight.innerHTML = "RESET";
    chronometer.stopClick();
  }
});

// Reset/Split Button
btnRight.addEventListener("click", () => {
  if (btnRight.classList.contains("reset")) {
    chronometer.resetClick();
  } else {
    chronometer.splitClick();
  }
});
```
-  Be careful, splitClick is not expecting a callback function, it does not need printTime at all. Also, you should make use of printSplit. So, inside your event listener, you should execute printSplit, instead of simply splitClick. Your event listener becomes:

```
btnRight.addEventListener("click", () => {
  if (btnRight.classList.contains("reset")) {
    chronometer.resetClick();
  } else {
    printSplit();
  }
});
```

Be careful, in printSplit you should add elements to the innerHTML, not setting it equal, otherwise you will only have the last split everytime:

```
function printSplit() {
  // ... your code goes here
  const currentSplit = chronometer.splitClick();
  splits.innerHTML += `<li>${currentSplit}</li>`;
}
```

- To make your reset button works perfectly, you need to printTime() once after you press the button reset, so that the time gets to 00;00 on the screen, and you need to clear your splits. Your event listener becomes:

```
// Reset/Split Button
btnRight.addEventListener("click", () => {
  if (btnRight.classList.contains("reset")) {
    chronometer.resetClick();
    printTime();
    clearSplits();
  } else {
    printSplit();
  }
});
```

PR #931, opened 2020-06-28
[REMOTE-WDPT-0602] - Raquel Nascimento, Cristina Matamoro

RaquelNascimento96:
- We've done both the iterations and the bonus - it's working as the demo
- However, our Jasmine is not all green - by introducing the callback() into startClick method, it returned an error
- Also, Jasmine is red on the last iteration since we added milliseconds, in a different format than expected


valmgisbert:
In your `callback()` you can have it as your `printMilliseconds()`; set a conditional to avoid getting errors if there is no extra argument. It's okay if the last one gives a failed test if you add the bonus; the tests aren't looking for it so they won't pass. Very well-structured and clean logic! Wonderful. In your 'click' eventListeners, you could do each conditional within the functions above instead of setting the innerHTML values within the if/else statements to have a more modular and cleaner code. Excellent job, girls! I am very proud of you!

PR #932, opened 2020-06-28
 [REMOTE-WDPT-0602] - Sunitha Mannam 

SunithaMannam:
lab complete with bonus

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,38 +1,41 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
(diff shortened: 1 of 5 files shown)


valmgisbert:
Remember to clean your console.logs. Great way to adapt your `startClick()` for milliseconds as well fro m the start without setting an extra `intervalID`! Very clear and concise code. You could also try practicing string interpolation where you can apply it. Excellent work!

PR #933, opened 2020-06-28
[Remote-PT-0620] Debora Crescenzo and Rodrigo Hernández

NubivagantYansa:
Done!


valmgisbert:
Very clear and organised logic! Very nice approach in using your `child` as a parameter for the `printSplit()` func. Also, love your use of `forEach`! I see some one-liners, look at you levelling up. Wonderful work, y'all!

PR #934, opened 2020-06-28
[REMOTE-WDPT-0602] - Alejandro Gonzalez, Matthew Howard

mattalanhoward:


mattalanhoward:
How can we fix the following issue?

The start/stop milliseconds works good the first time start/stop is clicked.  However, each consecutive time after the first click the milliseconds seems to double or triple in speed.  

Here is an example from the splits print out.  Notice how on the 11 second, the milliseconds restarts from zero.  Meaning it is going faster than the seconds.  
00:11:40
00:11:56
00:11:69
00:11:85
00:11:99
00:11:15
00:11:30

PR #942, opened 2020-07-15
[WDFT-REM-2007]Elise Dujardin

elisedjn:
Actually, I can't believe that I manage to already finish this lab.
For me the most challenging part was to understand the instructions, I did not find them really clear and even a bit confusing.For example, there are some hints that I don't understand why they are given... But anyway, my chronometer is working!

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,57 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
   startClick(callback) {
-    // ... your code goes here
+    console.log('startClick called');
+    this.intervalId = setInterval(() => { 
+      this.currentTime += 1;
+      callback();
+      console.log(this.currentTime);
+    }, 10)
   }
   getMinutes() {
-    // ... your code goes here
+    let minutes = Math.floor((this.currentTime/100) / 60);
+    return minutes;
   }
   getSeconds() {
-    // ... your code goes here
+    let seconds = Math.floor((this.currentTime/100) % 60);
+    return seconds;
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+  getMilliseconds() {
+    let milliseconds = this.currentTime % 100;
+    return milliseconds;
+  }
+  twoDigitsNumber(number) {
+    number = String(number);
+    if (number.length === 1 ) {
+      number = "0" + number;
+    }
+    return `${number}`;
   }
   stopClick() {
-    // ... your code goes here
+    console.log('stopClick called');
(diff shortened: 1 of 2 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,57 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
   startClick(callback) {
-    // ... your code goes here
+    console.log('startClick called');
```
IH-WebDev-TA-Remote, line 7: Just a small thing, but you may want to start removing console.logs when you submit labs or projects :)

javascript/chronometer.js
```diff
@@ -1,26 +1,57 @@
...
+    this.intervalId = setInterval(() => { 
+      this.currentTime += 1;
+      callback();
+      console.log(this.currentTime);
+    }, 10)
```
IH-WebDev-TA-Remote, line 12: Great! using the same setInterval for milliseconds, second and minutes.

javascript/index.js
```diff
@@ -14,52 +14,123 @@ let milUni = document.getElementById('milUni');
...
+  // Creating a new li with this value and adding it to the DOM
+  let newLi = document.createElement('li');
+  newLi.innerHTML = time;
+  splits.appendChild(newLi);
 }
 
 function clearSplits() {
-  // ... your code goes here
+  // looping while splits has child node and deleting every child node (=li)
+  while (splits.firstChild) {
```
IH-WebDev-TA-Remote, line 65: Very clever using the firstChild! :)

javascript/index.js
```diff
@@ -14,52 +14,123 @@ let milUni = document.getElementById('milUni');
...
 
 function setStopBtn() {
-  // ... your code goes here
+  // Changing the classes
+  btnLeft.classList.remove('start');
+  btnLeft.classList.add('stop');
+  // Changing the text
+  btnLeft.innerHTML = "STOP";
+  // Calling the startClick method
+  chronometer.startClick(printTime);
```
IH-WebDev-TA-Remote, line 77: For more coding clarity, it would make more sense that this startClick function call is in the addEventListener.

IH-WebDev-TA-Remote:
The reason why you might have found the instruction confusing is that there are many many ways to have a functioning chronometer, the instructions try to push you in one direction to be more precise with hints, but with a different logic these things might not make sense at first, which is totally understandable. Still we will pass on the feedback for more instruction clarity on this one :)
Now on to the solution... amazing work. It runs perfectly and the code is very clean and easy to understand. Good commenting is the best! This will specially come handy in your projects :)

PR #943, opened 2020-07-15
[WDFT-REM-2007] BRITTA HELM

brittahelm:
using the callback function was challenging, i kept getting confused on where to used () and where not but i got it to work eventually

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,56 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
+
   startClick(callback) {
-    // ... your code goes here
+    this.intervalId = setInterval( () => {
+      this.currentTime += 1;
+      callback();
+    }, 10)
   }
+
   getMinutes() {
-    // ... your code goes here
+    let minutes = ((this.currentTime/100) - (this.currentTime/100) % 60)/60;
+    return minutes;
   }
   getSeconds() {
-    // ... your code goes here
+    let seconds = Math.floor((this.currentTime/100)%60);
+    return seconds;
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+
+  getMilliseconds() {
+    let milliseconds = (this.currentTime % 100);
+    return milliseconds;
   }
+
+
+  twoDigitsNumber(num) {
+    let twoDigitNum = '';
+    if (num < 10) {
+      twoDigitNum = `0${num}`;
+    }
+    else {
+      twoDigitNum = `${num}`;
+    }
+    return twoDigitNum;
+  }
+
+
+
   stopClick() {
-    // ... your code goes here
+    clearInterval(this.intervalId);
   }
+
   resetClick() {
-    // ... your code goes here
+    this.currentTime = 0;
   }
+
   splitClick() {
-    // ... your code goes here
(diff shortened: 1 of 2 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,56 @@
...
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
+
   startClick(callback) {
-    // ... your code goes here
+    this.intervalId = setInterval( () => {
+      this.currentTime += 1;
+      callback();
+    }, 10)
```
IH-WebDev-TA-Remote, line 11: Good call using centiseconds so you can use it for the bonus as well :)

javascript/index.js
```diff
@@ -13,53 +13,69 @@ let milDec = document.getElementById('milDec');
...
+  printMinutes();
+  printSeconds();
+  printMilliseconds();
 }
 
 function printMinutes() {
-  // ... your code goes here
+  let minsFirstDig = chronometer.twoDigitsNumber(chronometer.getMinutes())[0];
+  let minsSecDig = chronometer.twoDigitsNumber(chronometer.getMinutes())[1];
+  document.querySelector('#minDec').innerText = minsFirstDig;
```
IH-WebDev-TA-Remote, line 26: In lines 4 to 14 you already have the reference of all the DOM elements you will be modifying, so you can target those variables instead of selecting the DOM elements again.

javascript/index.js
```diff
@@ -13,53 +13,69 @@ let milDec = document.getElementById('milDec');
...
-
-function setResetBtn() {
-  // ... your code goes here
-}
 
 // Start/Stop Button
 btnLeft.addEventListener('click', () => {
-  // ... your code goes here
+  if (btnLeft.classList.contains('start')) {
+    btnLeft.classList.replace('start', 'stop');
```
IH-WebDev-TA-Remote, line 48: To make your code more modular, you could create functions that change the classes of the button elements and you call those functions from here, I know it can be a little confusing at first but practicing this will help you when building more complex stuff.

javascript/index.js
```diff
@@ -13,53 +13,69 @@ let milDec = document.getElementById('milDec');
...
+    btnRight.classList.replace('reset', 'split');
+    btnLeft.innerText = 'STOP';
+    btnRight.innerText = 'SPLIT';
+    chronometer.startClick(printTime);
+  }
+  else {
+    btnLeft.classList.replace('stop', 'start');
+    btnRight.classList.replace('split', 'reset');
```
IH-WebDev-TA-Remote, line 56: very clever using the replace method! There are dozens of ways to change the classes but this is definitely a clever one :)

IH-WebDev-TA-Remote:
That part about callbacks can definitely be tricky and hard to understand, best advice for that is read your code line by line and think:
Do you want to get the function result at this point?
If yes, call the function. So, use the ( ).
If not, just pass it without the ( ) and you can invoke it later.
In this Lab, in startClick(printTime), The reason why you want to pass it as a callback parameter and not a invoke it, is because printTime updates the dom, and if you call it from index.js it will only update the DOM when you click the button, on the other hand, if you pass it as a callback to be invoked later, it will run inside the setInterval, meaning it will update the DOM every centisecond :)
Regarding the whole lab, amazing work. Everything works correct even with the bonus. I would say the most important feedback would be to reuse the DOM variables and try to make more functions to make your code more modular. But again, really good work. Keep it up!

PR #944, opened 2020-07-15
[WEB-REM-2007] RAFAEL DE LAZARO

rlaz133:
I'm mathematically challenged, so I had to use extra variables and processes to get the milliseconds.
Also, I later found out how to do the callback instead of executing the printTime from the load, but this is the solution I arrived at myself.

Definitely not an optimal use of memory, but happy that I managed to get it done.

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,39 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime =0;
+    this.intervalId=0;
+    this.intervalId2=0;
+    this.currentMilli=0;
   }
   startClick(callback) {
-    // ... your code goes here
+    this.intervalId = setInterval(()=>{this.currentTime++}, 1000)
+    this.intervalId2 = setInterval(()=>{this.currentMilli++}, 10)
+    console.log ('started')
   }
   getMinutes() {
-    // ... your code goes here
+    return Math.floor(this.currentTime/60)
   }
   getSeconds() {
-    // ... your code goes here
+    return Math.floor(this.currentTime%60)
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+  getMilliseconds() {
+    return Math.floor(this.currentMilli%100)
+  }
+
+  twoDigitsNumber(number) {
+    if (number>=10){return `${number}`}
+    else {return `0${number}`}
   }
   stopClick() {
-    // ... your code goes here
+    clearInterval(this.intervalId)
+    clearInterval(this.intervalId2)
+    console.log ('stopped')
   }
   resetClick() {
-    // ... your code goes here
+    this.currentTime=0;
+    this.currentMilli=0;
   }
   splitClick() {
-    // ... your code goes here
+    return `${this.twoDigitsNumber(this.getMinutes())}:${this.twoDigitsNumber(this.getSeconds())}:${this.twoDigitsNumber(this.getMilliseconds())}`
   }
 }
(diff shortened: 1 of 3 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,39 @@
...
-    // ... your code goes here
+    this.currentTime =0;
+    this.intervalId=0;
+    this.intervalId2=0;
+    this.currentMilli=0;
   }
   startClick(callback) {
-    // ... your code goes here
+    this.intervalId = setInterval(()=>{this.currentTime++}, 1000)
+    this.intervalId2 = setInterval(()=>{this.currentMilli++}, 10)
```
IH-WebDev-TA-Remote, line 10: There will be a lot of math involved, but instead of doing two setInvervals you could use one and work with it to get the bonus part. Even if you have to break Jasmine heh.

javascript/chronometer.js
```diff
@@ -1,26 +1,39 @@
...
+
+  twoDigitsNumber(number) {
+    if (number>=10){return `${number}`}
+    else {return `0${number}`}
   }
   stopClick() {
-    // ... your code goes here
+    clearInterval(this.intervalId)
+    clearInterval(this.intervalId2)
+    console.log ('stopped')
```
IH-WebDev-TA-Remote, line 30: At this point, remember to clean your code from console.logs. This will be specially helpful when submitting your projects.

javascript/chronometer.js
```diff
@@ -1,26 +1,39 @@
...
+    return Math.floor(this.currentTime%60)
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+  getMilliseconds() {
+    return Math.floor(this.currentMilli%100)
+  }
+
+  twoDigitsNumber(number) {
+    if (number>=10){return `${number}`}
```
IH-WebDev-TA-Remote, line 24: Just a little tip. When doing one line conditionals, you can remove the brackets and it will look better and still easy to read. Another tip would be for you to turn this into a ternary and start to get comfortable with those.

javascript/index.js
```diff
@@ -14,52 +14,63 @@ let milUni = document.getElementById('milUni');
 let splits = document.getElementById('splits');
 
 function printTime() {
-  // ... your code goes here
+  setInterval(()=>{
+  printMinutes(chronometer.twoDigitsNumber(chronometer.getMinutes()))
```
IH-WebDev-TA-Remote, line 18: This work but definitly not the best solution. The idea is to pass the callback that will run in the setInterval of the chronometer class. To do that you just want to pass the printTime function as a parameter inside the startClick call in line 60. and then Inside the startClick declaration just call the callback function.
Something like:
index.js Line 60: chronometer.stopClick(printTime) // the function is being passed to the chronometer object.
chronometer.js Line 9: callback() // the printTime function is being invoked inside the setInterval

javascript/index.js
```diff
@@ -14,52 +14,63 @@ let milUni = document.getElementById('milUni');
...
 }
 
 // Start/Stop Button
 btnLeft.addEventListener('click', () => {
-  // ... your code goes here
-});
+  btnLeft.classList.toggle('stop');
+  btnLeft.classList.toggle('start');
+  if (btnLeft.innerHTML === "START") {
+    btnLeft.innerHTML = "STOP";
```
IH-WebDev-TA-Remote, line 50: to make your code more modular, you could instead write the code that changes classes and then call those functions in the addEventListener.

IH-WebDev-TA-Remote:
Great work on the lab and good logic, specially coming out with that solution! However, working with callbacks will be essential later. So try to practice those and see exactly how they work as it will help you a lot along the way. Keep it up!

PR #946, opened 2020-07-15
WEB-REM-2007 Bodei

BodeiBrouwer:
I am not getting it to print. And the reset button is not working either.

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,39 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
-  }
+      this.currentTime = 0;
+      this.intervalId = 0;
+   }
+
   startClick(callback) {
-    // ... your code goes here
+    let intervalID = setInterval(() => {
+      this.currentTime++;
+    }, 1000)
   }
   getMinutes() {
-    // ... your code goes here
+    let minutes = this.currentTime/60
+    minutes = Math.floor(minutes)
+    return minutes
   }
   getSeconds() {
-    // ... your code goes here
+    let seconds = this.currentTime % 60
+    seconds = Math.floor(seconds)
+    return seconds
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+
+  twoDigitsNumber(number) {
+    if (number.toString().length < 2) {
+      number = "0" + number;
+    }
+    return number
   }
+
   stopClick() {
-    // ... your code goes here
+    clearInterval(this.intervalID);
   }
   resetClick() {
-    // ... your code goes here
+    this.currentTime = 0;
   }
   splitClick() {
-    // ... your code goes here
+    return `${this.twoDigitsNumber(this.getMinutes())}:${this.twoDigitsNumber(this.getSeconds())}`
   }
 }
--- a/javascript/index.js
+++ b/javascript/index.js
@@ -1,4 +1,4 @@
-const chronometer = new Chronometer();
+const chronometer = new Chronometer(0, 0);
 
 // get the buttons:
 const btnLeft = document.getElementById('btnLeft');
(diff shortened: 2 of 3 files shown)


BodeiBrouwer:
I've done a few updates. It prints sometimes but I have no more time to figure out what is going wrong. 
It still doesn't clear the split (don't have the right parent/child).
It does now print the actual time in the splits!

javascript/chronometer.js
```diff
@@ -1,26 +1,39 @@
...
   constructor() {
-    // ... your code goes here
-  }
+      this.currentTime = 0;
+      this.intervalId = 0;
+   }
+
   startClick(callback) {
-    // ... your code goes here
+    let intervalID = setInterval(() => {
```
IH-WebDev-TA-Remote, line 8: You are saving the interval ID into a simple variable, but not in the intervalID property (this.intervalId). So when you call the clearInterval in line 31, it is never clearing this interval, it is just trying to clear an interval whose id is 0.

javascript/chronometer.js
```diff
@@ -1,26 +1,39 @@
...
-    // ... your code goes here
+    clearInterval(this.intervalID);
   }
   resetClick() {
-    // ... your code goes here
+    this.currentTime = 0;
   }
   splitClick() {
-    // ... your code goes here
+    return `${this.twoDigitsNumber(this.getMinutes())}:${this.twoDigitsNumber(this.getSeconds())}`
```
IH-WebDev-TA-Remote, line 37: this works perfect, but to have a clearer code, you may want to save the function calls inside variables, and then return the interpolation of the variables. It is more lines of code but not always less lines is better.

javascript/index.js
```diff
@@ -14,15 +14,20 @@ let milUni = document.getElementById('milUni');
...
 function printTime() {
-  // ... your code goes here
+  printSeconds()
+  printMinutes() 
 }
 
 function printMinutes() {
-  // ... your code goes here
+  let minutes = chronometer.twoDigitsNumber(chronometer.getMinutes());
+  document.getElementById('minDec').innerHTML = minutes.charAt(0);
```
IH-WebDev-TA-Remote, line 23: You already have al DOM element references saved in variables in lines 4 to 14. You can use those variables instead of selecting again the DOM elements. It is better for performance.

javascript/index.js
```diff
@@ -31,35 +36,65 @@ function printMilliseconds() {
...
 }
 
 // Start/Stop Button
 btnLeft.addEventListener('click', () => {
-  // ... your code goes here
+ if (btnLeft == document.querySelector('.stop')) {
+  setStopBtn()
+ } else {
+  setStartBtn()
+  chronometer.startClick(printTime())
```
IH-WebDev-TA-Remote, line 89: You are executing the printTime when you call the startClick method (you are using the () ). This will print things only once, what you want to do instead, is pass the function as a paremeter, meaning you don't call the function. And then inside the chronometer class, you call the callback.
So it should look something like this:
index.js L 89: chronometer.startClick(printTime) // note you should not use ( ) when passing functions.
chronometer.js L10: callback() // then inside the chronometer, you call the function that was passed, inside the interval.

javascript/index.js
```diff
@@ -31,35 +36,65 @@ function printMilliseconds() {
...
+
+  const list = document.querySelector('#splits').parentNode
+  list.removeChild(list);
 }
 
 // Start/Stop Button
 btnLeft.addEventListener('click', () => {
-  // ... your code goes here
+ if (btnLeft == document.querySelector('.stop')) {
+  setStopBtn()
```
IH-WebDev-TA-Remote, line 86: same as you call the startClick function below, you need to call the stopClick function here.

javascript/index.js
```diff
@@ -31,35 +36,65 @@ function printMilliseconds() {
...
+  stopBtn.innerHTML = "STOP"
+
+  let resetBtn = document.getElementById('btnRight')
+  resetBtn.classList.toggle('split');
+  resetBtn.innerHTML = "SPLIT"
 }
 
 function setResetBtn() {
-  // ... your code goes here
+  chronometer.currentTime = 0;
```
IH-WebDev-TA-Remote, line 77: You already have a function inside of chronometer for clearing the currentTime, so it is better to just call that function, and you could do that in the addEventListener.

IH-WebDev-TA-Remote:
Good job on the lab, shows a lot of effort and just with some fixes things were showing on the DOM. Check the individual feedback lines and specially try to go over how the callbacks work. Let me know if you have a specific question or something that wasn't clear enough. Keep it up!

PR #947, opened 2020-07-15
[WDFT-REM-2007] Maëva Merrou

maevamerrou:


javascript/index.js
```diff
@@ -14,52 +15,122 @@ let milUni = document.getElementById('milUni');
...
+      btnLeft.innerText = "START";  
+    }
+  }
 }
 
+
+
 function setResetBtn() {
-  // ... your code goes here
+  btnLeft.addEventListener("click", function (){
```
IH-WebDev-TA-Remote, line 116: In these functions you are having some syntax errors that are breaking the code.
1. you are not closing the conditional of the if statement. Missing a ")" in Line 117.
2. you are not closing the addEventListener. Missing a ")" in Line 122.
This is happening in all functions.

javascript/index.js
```diff
@@ -14,52 +15,122 @@ let milUni = document.getElementById('milUni');
 let splits = document.getElementById('splits');
 
 function printTime() {
-  // ... your code goes here
+  getMinutes(); //return minutes
+  getSeconds()// return seconds
+  twoDigitsNumber(minutes)
```
IH-WebDev-TA-Remote, line 20: the minutes and seconds variables do not exist in the function. If you want to use the value of the getMinutes and getFunction you need to save those in the variables.

javascript/index.js
```diff
@@ -14,52 +15,122 @@ let milUni = document.getElementById('milUni');
 let splits = document.getElementById('splits');
 
 function printTime() {
-  // ... your code goes here
+  getMinutes(); //return minutes
```
IH-WebDev-TA-Remote, line 18: getMinutes and getSeconds are methods of the chronometer, so in order to call them you need to point to the chronometer. So: chronometer.getSeconds and chronometer.getMinutes

javascript/index.js
```diff
@@ -14,52 +15,122 @@ let milUni = document.getElementById('milUni');
...
+  getSeconds()// return seconds
+  twoDigitsNumber(minutes)
+  twoDigitsNumber(seconds)
+
+  minDec.innerHTML = minutes[0]
```
IH-WebDev-TA-Remote, line 23: you are already changing the DOM elements in the printMinutes and printSeconds functions. So instead of changing the same DOM elements in here, just call those functions.

javascript/index.js
```diff
@@ -14,52 +15,122 @@ let milUni = document.getElementById('milUni');
...
+    while (splits.firstChild) {
+      splits.removeChild(splits.firstChild)
+    }
+  }
 }
+  
 
 function setStopBtn() {
-  // ... your code goes here
+  btnLeft.addEventListener("click", function (){
```
IH-WebDev-TA-Remote, line 83: You already have an addEventListener at the bottom of the code, you want to run these functions that modify the DOM there. You want to have addEventListeners that call functions and not functions that call addEventListeners. Otherwise the functions are never being called.

IH-WebDev-TA-Remote:
Good job on the first part of the LAB, the logic for the class looks really good and easy to understand. However in the second part there is a lot to improve. I would recommend going through the solution and trying to understand how the addEventListeners are working. So try to go line by line on what is happening in the code everytime you click a button. I am sure you can do better! Keep it up! :)

PR #948, opened 2020-07-15
[WEB-REM-2007] Lanette Lanting

llanting:
I was struggling with this lab a bit... I tried making the counter work for half an hour, after which I saw that an open Jasmine-page prevented my IronChronometer-page from refreshing. Sometimes the learning points are not so much in the code itself as in using the tools around it! 

Speaking of tools, Google really was my best friend for this one. Eventually, I ended up doing to lab twice. One time trying to get it to work and Google a lot. Second time to really understand all the solutions I picked off the internet. That helped me a lot in understanding the lab and all the methods, the class, etc. And, it gave me the time to clean up my code a bit. 

Hardest things for me:
- The 'twoDigitsNumber' method: I first tried it with 'if (num < 9)', but couldn't get that to work - why is that?
- The resetClick(): it doesn't seem to work for me when calling it for the reset-button. I found a not-so-elegant solution

Looking forward to coding a game tomorrow!

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,71 @@
 class Chronometer {
   constructor() {
     // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
+
   startClick(callback) {
     // ... your code goes here
+    
+    this.intervalId = setInterval(() => {
+      this.currentTime++;
+      if (callback) {
+      callback();
+      }
+    }, 10)
   }
+
   getMinutes() {
     // ... your code goes here
+    return Math.floor((this.currentTime / 100) / 60);
   }
+
+
   getSeconds() {
     // ... your code goes here
+    return Math.floor((this.currentTime / 100) % 60);
+    
   }
-  twoDigitsNumber() {
+
+  getMiliSeconds() {
     // ... your code goes here
+    return Math.floor(this.currentTime % 100);
+    
   }
+
+  twoDigitsNumber(num) {
+    // ... your code goes here
+    
+    let n = num.toString();
+
+    if (n.length % 2 !== 0) {
+      return '0' + n;
+    }
+    else {
+      return n;
+    }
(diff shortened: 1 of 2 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,71 @@
...
+
   startClick(callback) {
     // ... your code goes here
+    
+    this.intervalId = setInterval(() => {
+      this.currentTime++;
+      if (callback) {
+      callback();
+      }
+    }, 10)
```
IH-WebDev-TA-Remote, line 16: Very nice work merging the bonus into the same setInterval!

javascript/chronometer.js
```diff
@@ -1,26 +1,71 @@
...
+  getMiliSeconds() {
     // ... your code goes here
+    return Math.floor(this.currentTime % 100);
+    
   }
+
+  twoDigitsNumber(num) {
+    // ... your code goes here
+    
+    let n = num.toString();
```
IH-WebDev-TA-Remote, line 40: remember to use better naming for your variables, for this case could be numAsStr or something similar.

javascript/index.js
```diff
@@ -15,51 +15,116 @@ let splits = document.getElementById('splits');
...
+    setResetBtn();
+    chronometer.stopClick();
+  }
 });
 
 // Reset/Split Button
 btnRight.addEventListener('click', () => {
   // ... your code goes here
+  if (btnRight.classList.contains('reset')) {
+    setStartBtn();
```
IH-WebDev-TA-Remote, line 106: The right button doesn't change the button classes, so calling this functions here is not needed. Same in L107, L126 and L127.

IH-WebDev-TA-Remote:
The more effort you put throughout the bootcamp the more you will gain from it :) 
The Lab solution is really good, everything is clear, well commented and runs perfect. 
Keep up the amazing work!

PR #949, opened 2020-07-15
[WEB-REM-2007] Marta Lourido

MartaLourido:
There are still some interactions to be completed.

The first interaction went well, I liked working with Jasmine again.

Then in the next ones, I got stuck and I spent hours trying to discover and searching on Google, but I didn't manage it.

In the end, it got late, and I have to sleep but I hope tomorrow I can complete it.

It is a little challenging, I want to study more to do this kind of lab without problem.

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,33 @@
 class Chronometer {
-  constructor() {
-    // ... your code goes here
+  constructor() {  //should not receive any arguments
+    this.currentTime = 0; //stablish the currentTime property
+    this.intervalId = 0; //Stablish the intervalID property
   }
-  startClick(callback) {
-    // ... your code goes here
-  }
+  startClick(callback) { //if every second currentTime increment by one, every 3 seconds it will increment by 3
+    let intervalId = setInterval(() => {
+			this.currentTime++;
+		}, 1000);
+	}
+   
   getMinutes() {
-    // ... your code goes here
-  }
+    return Math.floor(this.currentTime / 60);  //i used math.floor for return a number without decimals.
+    }
+
   getSeconds() {
-    // ... your code goes here
-  }
+    return Math.floor(this.currentTime % 60);  //i used math.floor for return a number without decimals.
+    }
+
   twoDigitsNumber() {
-    // ... your code goes here
+    return ("0" + this.currentTime).slice(-2);
   }
+
   stopClick() {
-    // ... your code goes here
+    return clearInterval (this.currentTime);
   }
   resetClick() {
-    // ... your code goes here
+    this.currentTime = 0;
   }
   splitClick() {
-    // ... your code goes here
-  }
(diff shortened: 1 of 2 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,33 @@
...
+    return clearInterval (this.currentTime);
   }
   resetClick() {
-    // ... your code goes here
+    this.currentTime = 0;
   }
   splitClick() {
-    // ... your code goes here
-  }
+    return `${this.twoDigitsNumber(this.getMinutes())}:${this.twoDigitsNumber(this.getSeconds())}`
```
IH-WebDev-TA-Remote, line 31: for a more clean and easy to understand code I would recommend that you save the function calls inside a variable and then use the variable inside the template literals. Less lines not always means a better code :)

javascript/chronometer.js
```diff
@@ -1,26 +1,33 @@
...
-  }
-  startClick(callback) {
-    // ... your code goes here
-  }
+  constructor() {  //should not receive any arguments
+    this.currentTime = 0; //stablish the currentTime property
+    this.intervalId = 0; //Stablish the intervalID property
+  }
+  startClick(callback) { //if every second currentTime increment by one, every 3 seconds it will increment by 3
+    let intervalId = setInterval(() => {
```
IH-WebDev-TA-Remote, line 7: You are saving the intervalId in a simple variable, but not in the class property (this.intervalId)

javascript/chronometer.js
```diff
@@ -1,26 +1,33 @@
...
+    }
+
   twoDigitsNumber() {
-    // ... your code goes here
+    return ("0" + this.currentTime).slice(-2);
   }
+
   stopClick() {
-    // ... your code goes here
+    return clearInterval (this.currentTime);
```
IH-WebDev-TA-Remote, line 25: this clearInterval won't work, what you want to pass inside the parenthesis is the intervalId, not the counter you have. The intervalId you have it saved in a variable in L7. If you save it in the property of the class, the you can easily use it here. (this.intervalId)

javascript/index.js
```diff
@@ -14,8 +14,12 @@ let milUni = document.getElementById('milUni');
 let splits = document.getElementById('splits');
 
 function printTime() {
-  // ... your code goes here
-}
+    let myInterval = setInterval(function() {
```
IH-WebDev-TA-Remote, line 17: The chronometer already has an interval running in it's class (maybe two in case of bonus) so you want to use those to print the time. That is why you can use call backs here. Create a function to update the DOM and then pass it as a callback to be run inside the class interval.

IH-WebDev-TA-Remote:
Good effort on the lab, however there were some issues with iteration 1, I know you are planning the project now but I would still suggest you go back to the lab and try it again, even checking the solution provided to see how things are functioning. Understanding this lab will help you a lot in the project, as there will be a lot of working with setinterval, DOM and callbacks in there.

PR #950, opened 2020-07-16
[WEB-REM-2020] ALVARO SANCHEZ LAMADRID

AlvaroSLam:
Iteration 1 completed, iteration 2 half of it

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,65 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
+
   startClick(callback) {
-    // ... your code goes here
+   this.currentTime = 0;
+   let intervalId = setInterval(() =>{
+     this.currentTime += 1;
+   }, 1000);  
+
+   return this.currentTime;
+
   }
+
   getMinutes() {
-    // ... your code goes here
+   let min = this.currentTime / 60;
+   min = Math.floor(min);
+   Number(min.toFixed(0));
+
+   return min
   }
+
+
   getSeconds() {
-    // ... your code goes here
+    if (this.currentTime === 0) {
+      return 0
+    }
+
+    let sec = this.currentTime % 60;
+
+    return sec;
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+
+  twoDigitsNumber(num) {
+    if (num > 9) {
+      return num;
+    } else {
+      return num +"0";
+    }
+
+
   }
+
   stopClick() {
-    // ... your code goes here
-  }
+    clearInterval(this.intervalId); 
+    }
+    
   resetClick() {
-    // ... your code goes here
+    this.currentTime = 0;
   }
-  splitClick() {
-    // ... your code goes here
+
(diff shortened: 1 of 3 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,65 @@
...
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
+
   startClick(callback) {
-    // ... your code goes here
+   this.currentTime = 0;
+   let intervalId = setInterval(() =>{
```
IH-WebDev-TA-Remote, line 9: you are saving the interval in a simple variable but you are not updating the class property, for that you need this.intervalId. The reason you do this is to later be able to clear it.

javascript/chronometer.js
```diff
@@ -1,26 +1,65 @@
...
+     this.currentTime += 1;
+   }, 1000);  
+
+   return this.currentTime;
```
IH-WebDev-TA-Remote, line 13: There is not need to return the currentTime, but this doesn't break anything. It is just not needed.

javascript/chronometer.js
```diff
@@ -1,26 +1,65 @@
...
+
   }
+
   getMinutes() {
-    // ... your code goes here
+   let min = this.currentTime / 60;
+   min = Math.floor(min);
+   Number(min.toFixed(0));
```
IH-WebDev-TA-Remote, line 20: Math.floor already gives you an integer number, so there is no need to use toFixed(0)

javascript/chronometer.js
```diff
@@ -1,26 +1,65 @@
...
+      return num +"0";
+    }
+
+
   }
+
   stopClick() {
-    // ... your code goes here
-  }
+    clearInterval(this.intervalId);
```
IH-WebDev-TA-Remote, line 47: This is doing nothing at the moment, because you are not passing the correct intervalId. See Line 9.

javascript/index.js
```diff
@@ -54,10 +54,21 @@ function setResetBtn() {
...
 // Start/Stop Button
-btnLeft.addEventListener('click', () => {
-  // ... your code goes here
-});
+btnLeft.addEventListener('click', () => {  
+  if (btnLeft.classList.contains("btn start")) {
+    btnLeft.classList.add("btn stop");
+    btnRight.classList.add("btn split")
+    console.log("Running")
+  } else if (!btnLeft.classList.contains("btn start")) {
```
IH-WebDev-TA-Remote, line 66: The logic was good but there were some issues.
Instead of: !btnLeft.classList.contains("btn start")
Do btnLeft.classList.contains("stop")
Also, you are adding the new classes (stop, split, start, reset) but you are not removing the previous ones, so you won't be able to change back and forth between them.

IH-WebDev-TA-Remote:
Good effort but there were some things breaking iteration 1 and the chronometer logic. Same for iteration 2. I know you are now working on the project but I would still recomment you try to finish at least until iteration 2 during the weekend. even side by side with the solution provided, just understanding everything that is happening in the code. This will help you a lot with the project as in there you will work a lot with DOM, intervals and callbacks.

PR #951, opened 2020-07-16
[WDFT-REM-2007]Gonzalo

gonzalo-development:


javascript/chronometer.js
```diff
@@ -1,26 +1,43 @@
...
-    // ... your code goes here
-  }
-  getSeconds() {
-    // ... your code goes here
-  }
-  twoDigitsNumber() {
-    // ... your code goes here
-  }
+
+this.intervalId = setInterval(() => this.currentTime++;
```
IH-WebDev-TA-Remote, line 9: If you need to run multiple lines inside a function you need to start with curly brackets. Missing "{" after the arrow function.

javascript/index.js
```diff
@@ -39,27 +41,60 @@ function clearSplits() {
...
+  }
+
+
 });
 
 // Reset/Split Button
 btnRight.addEventListener('click', () => {
-  // ... your code goes here
+
+  if()
```
IH-WebDev-TA-Remote, line 96: unfinished conditional was breaking code.

javascript/chronometer.js
```diff
@@ -1,26 +1,43 @@
...
+callback()
+}, 1000)
+
+}
+
+  getMinutes(currentTime) {
```
IH-WebDev-TA-Remote, line 15: These functions don't need a parameter currentTime, you are accesing the currentTime information via the this keyword (this.currentTime)

javascript/chronometer.js
```diff
@@ -1,26 +1,43 @@
...
+    return (Math.floor(this.currentTime/60))
+  }
+  getSeconds(currentTime) {
+    var minutes= this.getMinutes(currentTime)
+    
+    return (Math.floor(this.currentTime - (minutes * 60)))
+  }
+  twoDigitsNumber(currentTime) {
+    
+     currentTime =  "0" + currentTime
```
IH-WebDev-TA-Remote, line 25: You should not update the currentTime with a string. Doing this will cause the counter you have in the setInterval to break and start concatenating strings instead of adding 1 to the counter. Instead you just want to return the string for when the function is called later.

javascript/index.js
```diff
@@ -14,15 +14,17 @@ let milUni = document.getElementById('milUni');
...
 
 function printTime() {
-  // ... your code goes here
+  printMinutes()
+  printSeconds()
 }
 
 function printMinutes() {
-  // ... your code goes here
+  let mins= chronometer.getMinutes()
```
IH-WebDev-TA-Remote, line 22: mins is not the correct variable where the DOM element are. The variables are minDec and minUni (lines 8 and 9)
Also, getMinutes will throw you a number, you want a string instead and you get that by calling the twoDigitNumber

javascript/index.js
```diff
@@ -14,15 +14,17 @@ let milUni = document.getElementById('milUni');
...
 }
 
 function printSeconds() {
-  // ... your code goes here
+  let secs= chronometer.twoDigitsNumber(chronometer.getMinutes())
```
IH-WebDev-TA-Remote, line 26: same as above minutes comment.

javascript/index.js
```diff
@@ -39,27 +41,60 @@ function clearSplits() {
...
+  start.innerHTML= 'Reset'
+  start.className.add("reset");
+  start.className.remove("split")
+  
 }
 
 // Start/Stop Button
 btnLeft.addEventListener('click', () => {
-  // ... your code goes here
+  chronometer.startClick()
```
IH-WebDev-TA-Remote, line 78: You are starting the chronometer twice. One time in line 78 and a second time in line 82. This will definitely be an issue running the code.

javascript/index.js
```diff
@@ -39,27 +41,60 @@ function clearSplits() {
...
+  
+
+  if(!btnLeft.classList.contains('start')){
+    chronometer.startClick(printTime)
+    setStartBtn()
+    setResetBtn() 
+  }else{
+    setStopBtn()
+    setSplitBtn()
```
IH-WebDev-TA-Remote, line 87: Aside from changing the buttons, you also need to spop the timer here, this is essential to the code, otherwise you will have multiple setintervals running in the background. You do this by calling the chronometer stopClick function.

IH-WebDev-TA-Remote:
Some parts of the code look good but overall there are a lot of syntax errors and issues that are breaking everything. Check each individual feedback above. I would recommend you to go over the lab again with the solution and try to understand what is happening and how all the parts are functioning.

PR #956, opened 2020-07-18
Marcella - Amsterdam wd pt June2020

allecramarcella:
iteration 1 finished

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,46 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0
+    this.intervalId = 0
+   
   }
   startClick(callback) {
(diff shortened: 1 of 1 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,46 @@
...
   }
+
+
   getSeconds() {
-    // ... your code goes here
+    const seconds = this.currentTime - this.getMinutes() * 60
+    return seconds
   }
+
   twoDigitsNumber() {
```
guido-an, line 26: A shorter way to achieve the same result using a parameter and the ternary operator: 
` twoDigitsNumber (time) {`
   ` return time < 10 ? ${0}${time}` : `${time}`
`  }`

guido-an:
Well done!

PR #1013, opened 2020-09-23
[2009 rem] lovro hodak

LovroHodak:


javascript/chronometer.js
```diff
@@ -1,26 +1,63 @@
 class Chronometer {
   constructor() {
     // ... your code goes here
+    this.currentTime = 0
+    this.intervalId = 0
   }
   startClick(callback) {
     // ... your code goes here
+    setInterval(() => {
```
IH-WebDev-TA-Remote, line 9: VERY IMPORTANT. Every time you create a setInterval you need to save it in a variable, in this case in the intervalId attribute you created in the constructor. This essentially saved the ID of the interval running inside this variable, and it will be the only way to stop it later with the clearInterval function.

IH-WebDev-TA-Remote: Doing just this will make the functionality to stop the timer work.
`this.intervalId = setInterval(() => {...`

javascript/chronometer.js
```diff
@@ -1,26 +1,63 @@
...
   getSeconds() {
     // ... your code goes here
+    let seconds = this.currentTime % 60
+    return seconds
   }
-  twoDigitsNumber() {
+  twoDigitsNumber(input) {
     // ... your code goes here
+    if(input < 10){
+      return ('0' + input)
```
IH-WebDev-TA-Remote, line 28: no need for parenthesis here

javascript/chronometer.js
```diff
@@ -1,26 +1,63 @@
...
+    } else {
+      return (input.toString())
```
IH-WebDev-TA-Remote, line 30: also no need for outher parenthesis here

javascript/index.js
```diff
@@ -15,6 +15,22 @@ let splits = document.getElementById('splits');
 
 function printTime() {
```
IH-WebDev-TA-Remote, line 16: the functionality here works good but you already had some functions and method for this purpose. For example the methods getMinutes and getSeconds of the chronometer give you the current time in minutes and seconds, and the method twoDigitsNumber convert those numbers into an easier to read and use format. Here though, you are essentially redoing that logic instead of reusing those methods.

IH-WebDev-TA-Remote: See the solution we will share or the one explained in class to see the better (and simpler) approach.

javascript/index.js
```diff
@@ -40,26 +56,63 @@ function clearSplits() {
...
+  //chronometer.startClick(printTime)
+  //secUni.innerHTML = chronometer.currentTime
 });
 
 // Reset/Split Button
 btnRight.addEventListener('click', () => {
   // ... your code goes here
+  if(btnRight.innerText === 'RESET'){
+    //setSplitBtn()
+    chronometer.startClick(printTime)
```
IH-WebDev-TA-Remote, line 107: when the button reads RESET you want to run the chronometer method for reset. not start the chronometer again.

javascript/index.js
```diff
@@ -40,26 +56,63 @@ function clearSplits() {
...
+    //setStopBtn()
+  } else {
+    //setResetBtn()
+    //setStartBtn()
+    chronometer.stopClick()
```
IH-WebDev-TA-Remote, line 112: here you want to run the function that will create the splits. printSplit. not stop the chronometer.

IH-WebDev-TA-Remote:
Good job on the lab, it seems the issue with the intervalId was preventing everything from running and probably stopped you from moving ahead. Main feedback would be to review the way to use the methods created inside the Chronometer class and how to use the while updating the elements of the DOM, take a look at the solution and see how everything clicks together.

Great effort though, keep up the good work!

PR #1014, opened 2020-09-23
[2009] Aleksandra Tanasijenko

alekstana:
I finally did the first iteration and I think it works pretty well.
I managed to manipulate "stop and reset buttons" I think and added and event listener for "// Start/Stop Button", but I am not sure if is correct.
I will continue working on this Lab tomorrow!

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -4,33 +4,38 @@ class Chronometer {
     this.intervalId = 0;
 
   }
-  startClick(callback) {
-    setInterval(()=>{
-      this.currentTime++
-    },1000)
+  startClick() {
+    this.intervalId = setInterval(function(){ 
+    this.currentTime++; 
+  }, 1000);
+
  
   }
   getMinutes() {
     const minutes =  Math.floor(this.currentTime/60);
     return minutes;
+  }
 
-  }
   getSeconds() {
-    return this.currentTime%60;
+    const seconds = this.currentTime%60;
+    return seconds
   }
+
+
   twoDigitsNumber(elem) {
   //  return ('0' + elem).slice(-2)
+   
   if (elem<10){
-    elem = "0" +elem
+    elem = "0" + elem
   } else{
-    elem =elem 
+    elem = elem 
   }
   return elem
   }
 
 
  stopClick() {
-    clearInterval()
+    clearInterval(this.intervalId)
   }
   resetClick() {
     this.currentTime = 0
@@ -39,6 +44,9 @@ class Chronometer {
   splitClick() {
     let splitMinutes = this.twoDigitsNumber(this.getMinutes())
     let splitSeconds = this.twoDigitsNumber(this.getSeconds())
-    return splitMinutes + ":"+ splitSeconds
+    let splitValue = splitMinutes +`:`+ splitSeconds
+    return splitValue
   }
 }
+
+
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,44 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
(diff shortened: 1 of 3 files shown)


alekstana:
I tried to debug...i didn't succeed :(

javascript/chronometer.js
```diff
@@ -4,33 +4,38 @@ class Chronometer {
     this.intervalId = 0;
 
   }
-  startClick(callback) {
-    setInterval(()=>{
-      this.currentTime++
-    },1000)
+  startClick() {
+    this.intervalId = setInterval(function(){
```
IH-WebDev-TA-Remote, line 8: Interval function needs to be an arrow function in order for the this keyword to target to the correct element (the classChronometer)

javascript/index.js
```diff
@@ -9,20 +9,23 @@ let minDec = document.getElementById('minDec');
...
 
 function printTime() {
-  // ... your code goes here
+   printMinutes()
+  // function printSeconds()
 }
 
 function printMinutes() {
-  // ... your code goes here
+  minDec.innerHTML = Number(minutes[0])
```
IH-WebDev-TA-Remote, line 22: minutes doesn't exist in this code, the idea here was to use the methods available inside the chronometer element (getSeconds, getMinutes and twoDigitsNumber to get the needed results to update the DOM.

javascript/index.js
```diff
@@ -9,20 +9,23 @@ let minDec = document.getElementById('minDec');
...
+  minUni.innerHTML = Number(minutes[1])
 }
 
 function printSeconds() {
-  // ... your code goes here
+secUni.innerHTML = Number(seconds[0])
```
IH-WebDev-TA-Remote, line 27: Same as above minutes segment

javascript/index.js
```diff
@@ -31,35 +34,66 @@ function printMilliseconds() {
 }
 
 function printSplit() {
-  // ... your code goes here
+  btnRight.addEventListener('click', () => {
```
IH-WebDev-TA-Remote, line 37: you are already assigning the addEventListener to the buttons in the last part of the code, these will wait for the button to be clicked and then run this function. So in here, the only code needed is the one that prints the result. No need to create a new addEventListener after you click the button.

javascript/index.js
```diff
@@ -31,35 +34,66 @@ function printMilliseconds() {
...
+    if (btnRight.classList.contains('split')){
+    splits.innerHTML = splitValue
```
IH-WebDev-TA-Remote, line 39: Similar to the printMinutes and printSeconds comments, splitValue does not exist here, but you can use the chronometer splitClick method for this. you can call it simply doing: chronometer.splitClick()

javascript/index.js
```diff
@@ -31,35 +34,66 @@ function printMilliseconds() {
...
+  btnRight.innerHTML = "RESET"
 }
 
 // Start/Stop Button
 btnLeft.addEventListener('click', () => {
-  // ... your code goes here
+  if (btnLeft.classList.contains('start')){
+  setStopBtn()
+  setSplitBtn()
+  chronometer.startClick()
```
IH-WebDev-TA-Remote, line 73: The concept of callback and it's implementation was really hard to understand and apply, but it is essential for the code to work. Basically you have a function that updates the DOM (printTime) and you want that to be executed every second, so you want it to run inside the interval. So by passing the function as a parameter here and ten invoke the parameter inside the interval you will be able to achieve that.

IH-WebDev-TA-Remote:
Overall good effort, You finished all iteration 1 but it seems there were some issues preventing everything from updating in iteration 2, check all comments and also the solution we will send later. Also if something is not clear please let us know and we will explain it again :)

PR #1015, opened 2020-09-23
[REM-2009]zuzana

ZuzanaAK:
Hey Jorge, with some hints I have been able to complete the first iteration and a bit of the second one. I understand the stuff I have done. Unfortunately, I am feeling stuck on the second one. Would love to learn from the solutions! Thank you :)

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,39 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
+
   startClick(callback) {
-    // ... your code goes here
+    this.intervalId = setInterval(() => {
+    this.currentTime += 1
+  }, 1000);
   }
-  getMinutes() {
-    // ... your code goes here
+
+    getMinutes() {
+      let min = Math.floor(this.currentTime / 60);
+      return min;
+    }
+
+    getSeconds() {
+      let min = Math.floor(this.currentTime / 60);
+      let sec = this.currentTime - min * 60;
+      return sec;
+    }
+
+    twoDigitsNumber(digit) {
+
+       if (digit < 10) { 
+       return "0" + digit
+    }
   }
-  getSeconds() {
-    // ... your code goes here
+    stopClick() {
+      clearInterval(this.intervalId);
+    }
+    resetClick() {
+      this.currentTime = 0;
+    }
+    splitClick() {
+      return this.twoDigitsNumber(this.getMinutes()) + ":" + this.twoDigitsNumber(this.getSeconds())
+    }
   }
-  twoDigitsNumber() {
-    // ... your code goes here
-  }
-  stopClick() {
-    // ... your code goes here
-  }
-  resetClick() {
-    // ... your code goes here
-  }
-  splitClick() {
-    // ... your code goes here
-  }
-}
--- a/javascript/index.js
+++ b/javascript/index.js
@@ -14,7 +14,11 @@ let milUni = document.getElementById('milUni');
 let splits = document.getElementById('splits');
 
 function printTime() {
-  // ... your code goes here
+  let stop = document.querySelector("btn start")
+  stop.addEventListener("click", () => {
+    
+  
+})
 }
 
 function printMinutes() {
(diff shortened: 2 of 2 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,39 @@
...
+
+    getMinutes() {
+      let min = Math.floor(this.currentTime / 60);
+      return min;
+    }
+
+    getSeconds() {
+      let min = Math.floor(this.currentTime / 60);
+      let sec = this.currentTime - min * 60;
+      return sec;
```
IH-WebDev-TA-Remote, line 21: Interesting approach to solve this, but it works! :) A simpler one would be to use the % operator to get the remainder of the currentTime.
`this.currentTime % 60`

javascript/chronometer.js
```diff
@@ -1,26 +1,39 @@
...
+    }
+
+    twoDigitsNumber(digit) {
+
+       if (digit < 10) { 
+       return "0" + digit
+    }
```
IH-WebDev-TA-Remote, line 28: this condition will work only if the number has only one digit, but will not do anything if the number has two digits. The test will pass but you need to consider this option as well and just return the digit unchanged in that case.

javascript/chronometer.js
```diff
@@ -1,26 +1,39 @@
...
+  }
+    stopClick() {
+      clearInterval(this.intervalId);
+    }
+    resetClick() {
+      this.currentTime = 0;
+    }
+    splitClick() {
+      return this.twoDigitsNumber(this.getMinutes()) + ":" + this.twoDigitsNumber(this.getSeconds())
```
IH-WebDev-TA-Remote, line 37: Very nice solution!! and all in one line :)

javascript/index.js
```diff
@@ -40,23 +44,41 @@ function clearSplits() {
...
+  btnRight.className = "btn reset"
 }
 
 // Start/Stop Button
 btnLeft.addEventListener('click', () => {
   // ... your code goes here
+  if (btnLeft.innerText==="START") {
+    setStopBtn()
+    setSplitBtn()
+    //startClick() here??
```
IH-WebDev-TA-Remote, line 75: yes! this was the main part to conecting your two files, you want to invoke the method startClick() that starts your chronometer here. the way to do it is to access the chronometer object and run it's method. so:
`chronometer.startClick(printTime)`

IH-WebDev-TA-Remote: and why are we sending printTime as a parameter? this printTime function will be the main function that updates the DOM so it will change the numbers you see in the chronometer. So you want this to run not just one time, but every second. To do this you send the function (via parameter) as a callback to the startClick method and then inside the interval is when you want to execute the function.

javascript/chronometer.js
```diff
@@ -1,26 +1,39 @@
...
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
+
   startClick(callback) {
-    // ... your code goes here
+    this.intervalId = setInterval(() => {
+    this.currentTime += 1
```
IH-WebDev-TA-Remote, line 9: after this line you will want to execute the callback function.
`callback()`
This function will be the printTime function from index.js, just passed as a parameter. And you want to run it every second, so, inside the interval.

javascript/index.js
```diff
@@ -40,23 +44,41 @@ function clearSplits() {
...
+  } else { 
+    setStartBtn()
+    setResetBtn()
+    //stopClick() here??
```
IH-WebDev-TA-Remote, line 79: same as the startClick, yes, you want to run the one that stops it here. no need to use callbacks in this case though.

javascript/index.js
```diff
@@ -40,23 +44,41 @@ function clearSplits() {
...
+  }
+  setStopBtn()
```
IH-WebDev-TA-Remote, line 81: no need to run the function again here, this will cause problems.

javascript/index.js
```diff
@@ -14,7 +14,11 @@ let milUni = document.getElementById('milUni');
 let splits = document.getElementById('splits');
 
 function printTime() {
```
IH-WebDev-TA-Remote, line 16: no need to add more addEventListener in this functions. you only need them once for each button, and those are at the end of the code. This functions will not proceed to update the DOM. check the solutions on how these work up to this point.

IH-WebDev-TA-Remote:
Very good job on the first iteration and good effort on the second one, although you missed the segment that actually connects those two. Check the comments and the solutions we will send and try your best to understand what is happening, also if you have time go back to the second iteration now you know how to make the printTime function run and let us know if you need any help. Keep it up! :)

PR #1016, opened 2020-09-23
[REM - 2009] Vykintas Petrikas

VykiPetr:
I DID IT ALL! but i will probably forget how to explain it by morning.

I want to be the very best,
Like no one ever was.
To catch them is my real test,
To train them is my cause!
(I will travel across the land,
Searching far and wide.
Each Pokemon to understand
The power that's inside!)
Pokemon!
Gotta catch em' all!
It's you and me,
I know it's my destiny!
Pokemon!
Oh, you're my best friend,
In a world we must defend!
Pokemon!
Gotta catch em' all!
(A heart so true,
Our courage will pull us through!)
You teach me and I'll teach you,
Po-ke-mon!
Gotta catch em' all!
Gotta catch em' all!
Every challenge along the way,
With courage I will face!
I will battle every day,
To claim my rightful place!
Come with me, the time is right,
There's no better team!
Arm in arm, we'll win the fight,
It's always been our dream!
Pokemon!
Gotta catch em' all!
It's you and me,
I know it's my destiny!
Pokemon!
Oh, you're my best friend,
In a world we must defend!
Pokemon!
Gotta catch em' all!
(A heart so true,
Our courage will pull us through!)
You teach me and I'll teach you,
Po-ke-mon!
Gotta catch em' all!
Gotta catch em' all!
Gotta catch em' all!
Gotta catch em' all!
Gotta catch em' all!
Pokemon!
Gotta catch em' all!
It's you and me,
I know it's my destiny!
Pokemon!
Oh, you're my best friend,
In a world we must defend!
Pokemon!
Gotta catch em' all!
(A heart so true,
Our courage will pull us through!)
You teach me and I'll teach you,
Po-ke-mon!
Gotta catch em' all!
Gotta catch em' all!
Po-ke-mon!

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -5,7 +5,11 @@ class Chronometer {
   }
   startClick(callback) {
     this.intervalId = setInterval(() => {
-      callback()
+      if (callback) {
+      callback()}
+      if (this.currentTime > 100) {
+        this.currentTime = 0
+      }
       this.currentTime +=0.01
     }, 10)
   }
@@ -20,12 +24,18 @@ class Chronometer {
   getSeconds() {
     let time = this.currentTime
     let seconds = time%60
-    return seconds.toFixed([0])
+    return Math.floor(seconds)
   }
   getMiliseconds() {
     let time = this.currentTime
     let mil = time.toString()
-      return Math.floor(mil.substring(3,5))
+    console.log(mil)
+    console.log(time)
+    if (time < 10) {
+      return mil.substring(2,4) }
+      else if ( time > 10) {
+        return mil.substring(3,5)
+      }
   }
   twoDigitsNumber(number) {
     if (number < 10) {
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,49 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0
+    this.intervalId = ''
   }
   startClick(callback) {
-    // ... your code goes here
+    this.intervalId = setInterval(() => {
+      callback()
+      this.currentTime +=0.01
+    }, 10)
   }
   getMinutes() {
-    // ... your code goes here
+    let time = this.currentTime
+    let minutes = 0
+    if (time >= 60) {
+      minutes = time / 60
+    }
+    return Math.floor(minutes)
   }
   getSeconds() {
-    // ... your code goes here
+    let time = this.currentTime
+    let seconds = time%60
+    return seconds.toFixed([0])
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+  getMiliseconds() {
+    let time = this.currentTime
+    let mil = time.toString()
+      return Math.floor(mil.substring(3,5))
   }
-  stopClick() {
-    // ... your code goes here
+  twoDigitsNumber(number) {
+    if (number < 10) {
+      return ('0' + number)
+    } else {
+      return (number.toString())
+    }
+  }
+  stopClick() {  
+    clearInterval(this.intervalId)
   }
   resetClick() {
-    // ... your code goes here
+    this.currentTime = 0
   }
   splitClick() {
-    // ... your code goes here
+    let sec = this.twoDigitsNumber(this.getSeconds())
+    let min = this.twoDigitsNumber(this.getMinutes())
+    let mil = this.twoDigitsNumber(this.getMiliseconds())
+    return (min + ':' + sec + ':' + mil)
   }
 }
--- a/javascript/index.js
+++ b/javascript/index.js
@@ -1,5 +1,4 @@
 const chronometer = new Chronometer();
-
 // get the buttons:
 const btnLeft = document.getElementById('btnLeft');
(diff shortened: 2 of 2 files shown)


VykiPetr:
Well it doesnt work, but it looks like it does if you dont look carefully enough, my method of getting the milisecond to show on the split function is not fully foolproof.

VykiPetr:
And apparantely the jasmine tests went all red in the process of trying to make the miliseconds work, but I promise they were all green!

javascript/chronometer.js
```diff
@@ -1,26 +1,59 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0
+    this.intervalId = ''
```
IH-WebDev-TA-Remote, line 4: doesn't prevent the code from running or anything but the interval ID returned from the setInterval is always a number, to it makes sense you initialise it as a number instead of a string.

javascript/chronometer.js
```diff
@@ -1,26 +1,59 @@
...
   }
   startClick(callback) {
-    // ... your code goes here
+    this.intervalId = setInterval(() => {
+      if (callback) {
+      callback()}
+      if (this.currentTime > 100) {
+        this.currentTime = 0
+      }
```
IH-WebDev-TA-Remote, line 12: above conditional is preventing the currentTime from going pass 100, you might need this for time higher than 100 seconds.

javascript/chronometer.js
```diff
@@ -1,26 +1,59 @@
...
+      this.currentTime +=0.01
+    }, 10)
   }
   getMinutes() {
-    // ... your code goes here
+    let time = this.currentTime
+    let minutes = 0
+    if (time >= 60) {
+      minutes = time / 60
+    }
```
IH-WebDev-TA-Remote, line 21: this passes the test, but later on the feedback you will see that the idea is that this function returns the minutes even if the seconds are still lower than 60, so returning 0 minutes. This else condition will be needed.

javascript/chronometer.js
```diff
@@ -1,26 +1,59 @@
...
+    console.log(time)
+    if (time < 10) {
+      return mil.substring(2,4) }
+      else if ( time > 10) {
+        return mil.substring(3,5)
+      }
+  }
+  twoDigitsNumber(number) {
+    if (number < 10) {
+      return ('0' + number)
```
IH-WebDev-TA-Remote, line 42: no need for the outer parentheses here.

javascript/chronometer.js
```diff
@@ -1,26 +1,59 @@
...
+    } else {
+      return (number.toString())
```
IH-WebDev-TA-Remote, line 44: or here, you can remove the parentheses and it will still run.

javascript/index.js
```diff
@@ -14,52 +13,111 @@ let milUni = document.getElementById('milUni');
...
 function printTime() {
-  // ... your code goes here
+  milUni.innerHTML = Number(milUni.innerHTML) + 1
+  printMilliseconds()
+  printMinutes()
+  printSeconds()
 }
 
 function printMinutes() {
-  // ... your code goes here
```
IH-WebDev-TA-Remote, line 21: the code inside your functions printMinutes, printSeconds and printMilliseconds work fine, but the idea was to work around the methods of the chronometer that are already created. You have access to the functions getMinutes, getSeconds, getMilliseconds and even the method that turns them into a string form with two digits (twoDigitsNumber). So by calling those from the chronometer object you could have got the information needed to update the DOM without needed to write all the logic again.

javascript/index.js
```diff
@@ -14,52 +13,111 @@ let milUni = document.getElementById('milUni');
...
+    milDec.innerHTML = Number(milDec.innerHTML) + 1
+  } else if (Number(milDec.innerHTML) > 9) {
+    milDec.innerHTML = 0
+    secUni.innerHTML = Number(secUni.innerHTML) +1
+  }
 }
 
 function printSplit() {
-  // ... your code goes here
+    let splitList = document.querySelector('#splits')
```
IH-WebDev-TA-Remote, line 51: the lab already gives you this DOM element in line 13

javascript/index.js
```diff
@@ -14,52 +13,111 @@ let milUni = document.getElementById('milUni');
...
 }
 
 function clearSplits() {
-  // ... your code goes here
+  let splitList = document.querySelector('#splits')
+  let list = document.querySelector('#splits-container')
+  list.removeChild(splitList)
+  let ol = document.createElement('ol')
+  ol.setAttribute('id', 'splits')
+  list.appendChild(ol)
```
IH-WebDev-TA-Remote, line 64: this looks good but see below a simpler way to clear everything inside the split list ;) works like magic
`splits.innerHTML = "";`

IH-WebDev-TA-Remote:
Very good approach on iteration 1. Code looks great. For iteration 2 I see everything works great! even the bonus, however see the solution on how to work around the chronometer methods already created, right now you are making no use of them so even if you delete them the chronometer will still run. That means there is some code redundancy there.

Still, it works! and code looks good and good logic. So very well done! Now go out there and catch every pokémon ever.

PR #1017, opened 2020-09-23
[REM 2009] Ross Homaidan

ibirossi:
Very frustrating.  I dont understand why things dont work.  I dont understand how to get the functions in the class to show in the index.js...  :(

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,64 @@
 class Chronometer {
   constructor() {
+    this.currentTime = 0
+    this.intervalId = 0
+  
     // ... your code goes here
   }
   startClick(callback) {
+    
+    let intId = setInterval(()=>{
+      this.currentTime +=1
+    callback()
+  },1000) 
+  
+  
     // ... your code goes here
   }
   getMinutes() {
+    let minutes = 0
+    let time=this.currentTime
+    if (time>=60){
+    minutes=time/60
+    
+    }
+    return Math.floor(minutes)
     // ... your code goes here
   }
+  
   getSeconds() {
+    let seconds=this.currentTime
+    
+    return Math.floor(seconds%60)
+    
     // ... your code goes here
   }
-  twoDigitsNumber() {
+  twoDigitsNumber(time) {
+    if(time<10){
+      return (0+time.toString())
+    }
+    else return (time.toString())
+   
     // ... your code goes here
   }
   stopClick() {
+    clearInterval(this.intId)
+    
+
     // ... your code goes here
   }
   resetClick() {
+  this.currentTime=0
     // ... your code goes here
   }
+
   splitClick() {
-    // ... your code goes here
+    let splitSecs = this.twoDigitsNumber(this.getSeconds())
+    let splitMins = this.twoDigitsNumber(this.getMinutes())
+
+    return splitMins +":"+ splitSecs
+
+ 
+    
   }
 }
--- a/javascript/index.js
+++ b/javascript/index.js
@@ -14,10 +14,20 @@ let milUni = document.getElementById('milUni');
 let splits = document.getElementById('splits');
 
 function printTime() {
+  secUni.innerHTML = chronometer.getSeconds()
+
   // ... your code goes here
 }
 
 function printMinutes() {
+  //My idea is that I need to use double digit function and somehow get the first and second numbers. 
+  // I dont know how but for example, for minDec using [0] position of the 2 numbers
+  //and for the minUni the [1] position.  I dont know how!!
+
+  
+  
+    
+  
   // ... your code goes here
 }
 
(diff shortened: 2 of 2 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,64 @@
 class Chronometer {
   constructor() {
+    this.currentTime = 0
+    this.intervalId = 0
+  
     // ... your code goes here
   }
   startClick(callback) {
+    
+    let intId = setInterval(()=>{
```
IH-WebDev-TA-Remote, line 10: you created an attribute for the intervalId in line 4, so it is better to save the interval id inside that attribute instead of a simple variable, the reason is that this variable only exists inside this scope, so you won't be able to access if later to clear it.
`this.intervalId = setInterval(()=>{...`

javascript/chronometer.js
```diff
@@ -1,26 +1,64 @@
...
+  twoDigitsNumber(time) {
+    if(time<10){
+      return (0+time.toString())
+    }
+    else return (time.toString())
+   
     // ... your code goes here
   }
   stopClick() {
+    clearInterval(this.intId)
```
IH-WebDev-TA-Remote, line 45: there is no intId attribute, the name is intervalId. by correcting this and the comment on line 10, the stop functionality will now work

javascript/chronometer.js
```diff
@@ -1,26 +1,64 @@
...
+  
     // ... your code goes here
   }
   getMinutes() {
+    let minutes = 0
+    let time=this.currentTime
+    if (time>=60){
+    minutes=time/60
+    
+    }
```
IH-WebDev-TA-Remote, line 24: even if Jasmine is fine with it you are also expected to return a number if the time is lower than 60, so you can create an else conditional or even remove the conditional altogether and divide the time between 60.

javascript/chronometer.js
```diff
@@ -1,26 +1,64 @@
...
+    let seconds=this.currentTime
+    
+    return Math.floor(seconds%60)
+    
     // ... your code goes here
   }
-  twoDigitsNumber() {
+  twoDigitsNumber(time) {
+    if(time<10){
+      return (0+time.toString())
```
IH-WebDev-TA-Remote, line 38: no need for the outer parenthesis

javascript/chronometer.js
```diff
@@ -1,26 +1,64 @@
...
+    }
+    else return (time.toString())
```
IH-WebDev-TA-Remote, line 40: same here, you can remove the outer parenthesis

javascript/index.js
```diff
@@ -14,10 +14,20 @@ let milUni = document.getElementById('milUni');
 let splits = document.getElementById('splits');
 
 function printTime() {
+  secUni.innerHTML = chronometer.getSeconds()
```
IH-WebDev-TA-Remote, line 17: good approach, this helps you check that the DOM is being updated every second! However the idea of this function is that it is passed as a callback, and it will call the other functions that will update the DOM (printMinutes and printSeconds)

javascript/index.js
```diff
@@ -14,10 +14,20 @@ let milUni = document.getElementById('milUni');
...
+
   // ... your code goes here
 }
 
 function printMinutes() {
+  //My idea is that I need to use double digit function and somehow get the first and second numbers. 
+  // I dont know how but for example, for minDec using [0] position of the 2 numbers
+  //and for the minUni the [1] position.  I dont know how!!
```
IH-WebDev-TA-Remote, line 25: yes! using bracket notations was the correct approach! you just needed to see how to access those values. you know that you have access to the chronometer object (see line 1) that holds all the attributes and methods you created on chronometer.js. So what you need to do is invoke those methods and use the results. for Example:
`chronometer.getSeconds()` will get you the current seconds as a number
`chronometer.twoDigitsNumber()` will transform any number you pass as a two digit string
so...
`chronometer.twoDigitsNumber(chronometer.getMinutes())` 
... will get you the current seconds as a string and you could apply bracket notation to it like you mentioned.

javascript/index.js
```diff
@@ -35,31 +45,71 @@ function printSplit() {
...
 
 function clearSplits() {
+ 
   // ... your code goes here
 }
 
 function setStopBtn() {
+
+  btnLeft.innerHTML ='STOP'
+  btnLeft.setAttribute = ('class', 'btn stop')// wont change color, but split does. why?
```
IH-WebDev-TA-Remote, line 55: when working around classes for DOM elements it is better to use classList or className.

` btnLeft.className = "btn stop"`

or ...

`btnLeft.classList.toggle("stop")`
`btnLeft.classList.toggle("start")`

IH-WebDev-TA-Remote:
Good job in iteration 1 and good start on 2! seems you were having some difficulties on the way to use the chronometer methods but were on the right track. Check the solution we will send later to see how you can access them to update DOM.

Still things were updating on the chronometer and the buttons were working! Keep it up!

PR #1018, opened 2020-09-23
[ REM 2009 ] Jordane Ruiz

jordaneruiz:


javascript/chronometer.js
```diff
@@ -1,26 +1,47 @@
...
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0; 
+    this.intervalId = 0; 
   }
   startClick(callback) {
-    // ... your code goes here
+    this.intervalId = setInterval(() => {
+    this.currentTime += 1
+    // this is where i want to invoke the code //this callback will update the DOM
```
IH-WebDev-TA-Remote, line 9: just as the comment, yes, this is where you needed to invoke the callback function. simply adding: callback()

javascript/chronometer.js
```diff
@@ -1,26 +1,47 @@
...
   }
+
   resetClick() {
-    // ... your code goes here
+    this.currentTime = 0;
   }
+
   splitClick() {
-    // ... your code goes here
+    return `${this.twoDigitsNumber(this.getMinutes())}:${this.twoDigitsNumber(this.getSeconds())}`
```
IH-WebDev-TA-Remote, line 45: very good solution here, all in a single line!

javascript/index.js
```diff
@@ -39,27 +48,51 @@ function clearSplits() {
...
-  // ... your code goes here
+  btnRight.innerText = "RESET"
+  btnRight.className = "btn reset"
 }
 
 // Start/Stop Button
 btnLeft.addEventListener('click', () => {
-  // ... your code goes here
+  if (btnLeft.innerText == "START") {
+    chronometer.startClick();
```
IH-WebDev-TA-Remote, line 73: this method will now expect a parameter that will be invoked inside the interval, in this case it is the printTime function that updates the DOM, the reason is: you want to continuously update the DOM every time the chronometer timer updates.

javascript/index.js
```diff
@@ -14,15 +14,24 @@ let milUni = document.getElementById('milUni');
 let splits = document.getElementById('splits');
 
 function printTime() {
-  // ... your code goes here
+  getMinutes()
+  getSeconds()
```
IH-WebDev-TA-Remote, line 18: getMinutes and getSeconds are methods of the chronometer, the functions that you wanted to call here are printMinutes and printSeconds, that are running the above methods.

javascript/index.js
```diff
@@ -14,15 +14,24 @@ let milUni = document.getElementById('milUni');
...
 }
 
 function printMinutes() {
-  // ... your code goes here
+  let min = chronometer.twoDigitsNumber(chronometer.getMinutes());
+  let minSplit = min.split('');
```
IH-WebDev-TA-Remote, line 23: no need to split the string into an array to use bracket notation, you can also use bracket notation on strings :)

IH-WebDev-TA-Remote:
Iteration 1 is very good! the tests passed and the code is good and easy to read.
For the second you just needed to pass the callback, invoke it and change the functions being called in printTime, but the rest of the code was good.

Keep up the good work! :)

PR #1019, opened 2020-09-23
[REM-2009] Sumiya Ushiro

sumi0820:
I was stuck on the iteration 1 as it was still difficult for me to debug the issues relaying on Jasmine, and the logic basically gives outputs when it is run, so it is hard to see what would be happening at once, at least to me. 

On the other hands, the iteration 2 and 3  are still understandable. Although I did not manage to use twoDigitsNumber() function in Chronometer.js, I could display digits, but the codes in index.js got dirty and redundant, which I admit that is not a good approach...

The bonus part, I figured out how to display milliseconds, but the problem that I was not able to solve was duplicate digits. Somehow, when it reaches 99, getMilli() function gives 99 twice even though there is if condition to filter out and currentMilTime increases 1 by 1. Due to this, the timer looks like it stopped for a moment between 99 to 00. 

Overall, I re-recognized my weaknesses are those logic creation/manipulation that are run/to be run at backend. I have confidence to manipulate HTML and CSS as I can see what is going on there, but it will take time to get used to creating the logics more efficiently.

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,67 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0;
+    this.currentMilTime = 0;
+    this.intervalMilId = 0;
   }
+
   startClick(callback) {
-    // ... your code goes here
+    this.intervalId = setInterval(() => {
+      this.currentTime++;
+      callback();
+    }, 1000);
+    this.intervalMilId = setInterval(() => {
+      this.currentMilTime++;
+
+      callback();
+    }, 10);
   }
+
   getMinutes() {
-    // ... your code goes here
+    let min = Math.floor(this.currentTime / 60);
+    return min;
   }
+
   getSeconds() {
-    // ... your code goes here
+    let sec = this.currentTime % 60;
+
+    return sec;
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+
+  twoDigitsNumber(time) {
+    return time.toString().length == 1 ? `0${time}` : time.toString();
   }
   stopClick() {
-    // ... your code goes here
+    clearInterval(this.intervalId);
+    clearInterval(this.intervalMilId);
   }
   resetClick() {
-    // ... your code goes here
+    this.currentTime = 0;
   }
-  splitClick() {
-    // ... your code goes here
+  splitClick(time) {
+    let min = this.getMinutes();
+    let sec = this.getSeconds();
+    let mil = this.getMil();
+
+    if (min > 10 || sec > 10) {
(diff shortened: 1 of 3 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,67 @@
...
-    // ... your code goes here
+    let sec = this.currentTime % 60;
+
+    return sec;
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+
+  twoDigitsNumber(time) {
+    return time.toString().length == 1 ? `0${time}` : time.toString();
```
IH-WebDev-TA-Remote, line 33: nice! using ternaries like a pro dev!

javascript/chronometer.js
```diff
@@ -1,26 +1,67 @@
...
+    clearInterval(this.intervalId);
+    clearInterval(this.intervalMilId);
   }
   resetClick() {
-    // ... your code goes here
+    this.currentTime = 0;
   }
-  splitClick() {
-    // ... your code goes here
+  splitClick(time) {
```
IH-WebDev-TA-Remote, line 42: splitClick is not expecting a parameter. Also you are not using it.

javascript/index.js
```diff
@@ -1,65 +1,138 @@
...
+    let secString = String(sec).split("");
+    console.log(secString);
+    secUni.innerText = secString[1];
+    secDec.innerText = secString[0];
+  } else if (sec == 0) {
+    secUni.innerText = sec;
+    secDec.innerText = 0;
+  } else {
+    secUni.innerText = sec;
+  }
```
IH-WebDev-TA-Remote, line 48: this code works good but could be simplified, mostly by using the twoDigitsNumber method.

javascript/index.js
```diff
@@ -1,65 +1,138 @@
...
+  ol.appendChild(li);
 }
 
 function clearSplits() {
-  // ... your code goes here
+  let list = document.querySelectorAll(".split-item");
+
+  list.forEach((item) => {
+    item.parentNode.removeChild(item);
+  });
```
IH-WebDev-TA-Remote, line 81: there is a simpler approach for this
`splits.innerHTML = "";`

javascript/index.js
```diff
@@ -1,65 +1,138 @@
...
+
+  chronometer.currentTime = 0;
+  chronometer.currentMilTime = 0;
```
IH-WebDev-TA-Remote, line 84: this works as well but you have a method of the chronometer to for the reset click.

IH-WebDev-TA-Remote:
The solution is good, and everything is working! even the bonus, definitely the main feedback is working more with the functions already created in order to avoid code redundancy, but even you yourself realised that :)

Overall though, amazing work, keep it up!

PR #1020, opened 2020-09-24
[REM-2009] Santiago Andrés Ardiles Páez

SantAndresP:
I apologise for taking so long, but I've never worked with multiple files before and things were breaking all over the place.

``if ((minutes < 10 || seconds < 10) && milliseconds > 9) {``
``return `${0}${minutes}:${0}${seconds}:${milliseconds}`;``
``} else if ((minutes < 10 || seconds < 10) && milliseconds < 9) {``
``  return `${0}${minutes}:${0}${seconds}:${0}${milliseconds}`;``
``} else {``
``  return `${minutes}:${seconds}:${milliseconds}`;``
``}``

I know that looks horrible, but I couldn't find another way.

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,72 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.current_seconds = 0;
+    this.interval_seconds = 0;
+
+    this.current_milliseconds = 0;
+    this.interval_milliseconds = 0;
   }
-  startClick(callback) {
-    // ... your code goes here
+
+  startTimer(callback) {
+    this.startMilliseconds(callback);
+    this.startSeconds(callback);
   }
+
+  startSeconds(callback) {
+    this.interval_seconds = setInterval(() => {
+      this.current_seconds++;
+      callback();
+    }, 1000);
+  }
+
+  startMilliseconds(callback) {
+    this.interval_milliseconds = setInterval(() => {
+      this.current_milliseconds++;
+      callback();
+    }, 1);
+  }
+
   getMinutes() {
-    // ... your code goes here
+    return Math.floor(this.current_seconds / 60);
   }
+
   getSeconds() {
-    // ... your code goes here
+    return this.current_seconds - this.getMinutes() * 60;
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+
+  getMilliseconds() {
+    return this.current_milliseconds % 100;
   }
+
+  twoDigitsNumber(current_time) {
+    if (current_time < 10) {
+      return "0" + String(current_time);
+    } else {
+      return String(current_time);
+    }
+  }
+
   stopClick() {
-    // ... your code goes here
+    clearInterval(this.interval_milliseconds);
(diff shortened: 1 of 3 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,72 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.current_seconds = 0;
```
IH-WebDev-TA-Remote, line 3: Best practices, remember to use camelCasing for your JS variables.

javascript/chronometer.js
```diff
@@ -1,26 +1,72 @@
...
+
+    this.current_milliseconds = 0;
+    this.interval_milliseconds = 0;
   }
-  startClick(callback) {
-    // ... your code goes here
+
+  startTimer(callback) {
+    this.startMilliseconds(callback);
+    this.startSeconds(callback);
```
IH-WebDev-TA-Remote, line 12: good approach for the bonus and continuing working with callbacks, nice!

javascript/chronometer.js
```diff
@@ -1,26 +1,72 @@
...
+
+  getMilliseconds() {
+    return this.current_milliseconds % 100;
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+
+  twoDigitsNumber(current_time) {
+    if (current_time < 10) {
+      return "0" + String(current_time);
```
IH-WebDev-TA-Remote, line 43: current_time will be automatically converted to a string here when you concatenate it with another string. so no need for the String( )

javascript/chronometer.js
```diff
@@ -1,26 +1,72 @@
...
+    const seconds = this.getSeconds();
+    const milliseconds = this.getMilliseconds();
+
+    if ((minutes < 10 || seconds < 10) && milliseconds > 9) {
+      return `${0}${minutes}:${0}${seconds}:${milliseconds}`;
+    } else if ((minutes < 10 || seconds < 10) && milliseconds < 9) {
+      return `${0}${minutes}:${0}${seconds}:${0}${milliseconds}`;
+    } else {
+      return `${minutes}:${seconds}:${milliseconds}`;
+    }
```
IH-WebDev-TA-Remote, line 70: the conditional above works good but you already have a method that converts the number into a two digit string, so this code is a little redundant. Instead you could have use that.

javascript/index.js
```diff
@@ -1,65 +1,108 @@
...
-  // ... your code goes here
+  let milliseconds = chronometer.getMilliseconds();
+  milliseconds = chronometer.twoDigitsNumber(milliseconds);
+
+  return milliseconds
 }
 
 function printSplit() {
-  // ... your code goes here
+  const splits_list = document.querySelector("#splits");
```
IH-WebDev-TA-Remote, line 53: this DOM element variable is already accessed in Line 14.

IH-WebDev-TA-Remote:
It doesn't look that bad but definitely reusing the  methods of the chronometer would have been better. Still overall everything is working and the code looks good. Keep it up!

PR #1021, opened 2020-09-24
rem 2009 katie stamey

kmstamey:


javascript/chronometer.js
```diff
@@ -1,26 +1,47 @@
...
+    this.currentTime = 0;
+    this.intervalId = 0;
     // ... your code goes here
   }
   startClick(callback) {
-    // ... your code goes here
+    this.intervalId = setInterval(() => {
+      this.currentTime++;
+      if (callback) {
+      callback();
```
IH-WebDev-TA-Remote, line 11: very clever to use the condition that will evaluate if it is not undefined first.

javascript/chronometer.js
```diff
@@ -1,26 +1,47 @@
...
+  // ... your code goes here
   }
-  twoDigitsNumber() {
+  twoDigitsNumber(digit) {
     // ... your code goes here
+    if (digit < 10) { 
+      return "0" + digit;
+    }
+ 
+    return digit;
```
IH-WebDev-TA-Remote, line 30: right now it returns a number in this line, this will cause issues when invoking the function in index.js as it will try to work around a string.

javascript/index.js
```diff
@@ -1,30 +1,39 @@
...
+  //callbacks of startClick prints the time, so you have to pass this when calling the startClick function
+
+  printMinutes();
+  printSeconds();
+  printMilliseconds();
 }
 
 function printMinutes() {
   // ... your code goes here
+  let min = chronometer.twoDigitsNumber(chronometer.getMinutes());
```
IH-WebDev-TA-Remote, line 24: twoDigitsNumber will return a number if the number has two digits, so you won't be able to use bracket notation on numbers.

javascript/chronometer.js
IH-WebDev-TA-Remote: `return digit.toString()` would fix this

javascript/index.js
```diff
@@ -39,27 +48,46 @@ function clearSplits() {
...
 
 // Start/Stop Button
 btnLeft.addEventListener('click', () => {
   // ... your code goes here
+
+  // alert('Hello Katie');
+  if (btnLeft.classList.contains('start') === true) {
+    chronometer.startClick(() => {
+      printTime();
+    });
```
IH-WebDev-TA-Remote, line 78: very clever using an anonymous function that will run printTime, however, just passing printTime as an argument would work just as well.

IH-WebDev-TA-Remote:
Very nice! the first iteration works great and you managed to link them both and make the chronometer work to some extent, it seems there was a small issue preventing you to continue though. Code looks clean and easy to read. Keep it up!

PR #1022, opened 2020-09-24
[REM - 2009] Mohammed imran

14imran:
1 iteration is completed but second iteration not done. I was not able to break it down.

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,63 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+  // ... your code goes here
+ 
+this.currentTime = 0;
+this.intervalId = 0;
+
   }
+
+
   startClick(callback) {
     // ... your code goes here
+    setInterval(() => {
+      callback();
+      this.currentTime += 1;
+    },1000);
   }
   getMinutes() {
     // ... your code goes here
+
+let minutes = 0;
+let time = this.currentTime;
+if(time >= 60){
+  minutes = (time/ 60);
+}
+return Math.floor(minutes);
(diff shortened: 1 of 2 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,63 @@
...
     // ... your code goes here
+    setInterval(() => {
+      callback();
+      this.currentTime += 1;
+    },1000);
   }
   getMinutes() {
     // ... your code goes here
+
+let minutes = 0;
```
IH-WebDev-TA-Remote, line 21: check indentation of elements here, they should look inside the function for readability.

javascript/chronometer.js
```diff
@@ -1,26 +1,63 @@
...
+    // let sec = Math.abs(this.currentTime)
+    return sec;
   }
-  twoDigitsNumber() {
+  twoDigitsNumber(input) {
     // ... your code goes here
+    if(input<10)
+    {
+      return "0" + input;
+    }
```
IH-WebDev-TA-Remote, line 40: check the correct formatting of if conditionals, also some spacing for the condition inside like this:

```
if (input < 10) {
      return "0" + input;
    }
```

javascript/chronometer.js
14imran: okay. i will check it.

IH-WebDev-TA-Remote:
Good solution for iteration 1 and I also see you managed to link them and run the chronometer, but still a lot of code there missing for iteration 2. Go back to it even with the solution to see how the DOM manipulation part and use of methods is done. Keep it up!

PR #1023, opened 2020-09-24
[REM 2009]Marta Obregon

MartaObregon:
not printing time :(

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,39 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
-  startClick(callback) {
-    // ... your code goes here
+  startClick(printTime) {
+    setInterval(() =>{
+        this.currentTime++
+        printTime(this.currentTime)
+      },1000)
   }
   getMinutes() {
-    // ... your code goes here
+    let mins = Math.floor(this.currentTime / 60);
+    return mins
   }
+
   getSeconds() {
-    // ... your code goes here
+    let secs = (this.currentTime%60)
+    return secs
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+  twoDigitsNumber(number) {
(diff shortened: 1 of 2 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,39 @@
...
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
-  startClick(callback) {
-    // ... your code goes here
+  startClick(printTime) {
+    setInterval(() =>{
+        this.currentTime++
+        printTime(this.currentTime)
```
IH-WebDev-TA-Remote, line 9: the printTime function should not have any parameters.

javascript/index.js
```diff
@@ -13,16 +13,27 @@ let milDec = document.getElementById('milDec');
 let milUni = document.getElementById('milUni');
 let splits = document.getElementById('splits');
 
-function printTime() {
-  // ... your code goes here
+function printTime(minutes, seconds) {
+  printMinutes(minutes)
+  printSeconds(seconds)
```
IH-WebDev-TA-Remote, line 18: same here, printTime function should not have any parameters, same for the printMinutes and printSeconds

javascript/index.js
```diff
@@ -13,16 +13,27 @@ let milDec = document.getElementById('milDec');
...
+
+
 }
 
-function printMinutes() {
-  // ... your code goes here
+function printMinutes(minutes) {
+ chronometer.twoDigitsNumber(chronometer.getMinutes(minutes))
```
IH-WebDev-TA-Remote, line 24: getMinutes should not have any parameters as well. You can see that inside the chronometer it works with the currentTime variable. So it doesn't need any variable passed via parameter.

IH-WebDev-TA-Remote:
Good work on iteration 1 and good approach on two, the main issue you were having was trying to pass information via parameters when it was not needed. Check the solution if you have time (right now project is priority) to see how was that handled, keep it up! :)

PR #1024, opened 2020-09-24
[REM-2009] Juan Jose

JJCExcelsior:


javascript/chronometer.js
```diff
@@ -1,26 +1,40 @@
...
   }
   getMinutes() {
-    // ... your code goes here
+    let currentMinutes = Math.floor(this.currentTime / 60);
+    // console.log(currentMinutes)
+    return currentMinutes;
   }
   getSeconds() {
-    // ... your code goes here
+    let currentSeconds = (this.currentTime % 60);
```
IH-WebDev-TA-Remote, line 17: no need for parenthesis here

javascript/chronometer.js
```diff
@@ -1,26 +1,40 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
-  startClick(callback) {
-    // ... your code goes here
+  startClick(printTime) {
+    this.intervalId = setInterval(() => {
```
IH-WebDev-TA-Remote, line 7: VERY IMPORTANT. Every time you create a setInterval you need to save it in a variable, in this case in the intervalId attribute you created in the constructor. This essentially saved the ID of the interval running inside this variable, and it will be the only way to stop it later with the clearInterval function.

javascript/chronometer.js
```diff
@@ -1,26 +1,40 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
-  startClick(callback) {
-    // ... your code goes here
+  startClick(printTime) {
```
IH-WebDev-TA-Remote, line 6: the callback function printTime needed to be called inside the interval, as it is mentioned in the instructions.

javascript/chronometer.js
```diff
@@ -1,26 +1,40 @@
...
   getSeconds() {
-    // ... your code goes here
+    let currentSeconds = (this.currentTime % 60);
+    return currentSeconds;
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+  twoDigitsNumber(number) {
+    if (number < 10) {
+     return ("0" + number)
```
IH-WebDev-TA-Remote, line 22: no need to use outer parenthesis here either

IH-WebDev-TA-Remote:
the code itself looks perfect for iterations 1 and 2, however the middle section that connects the two and adds the eventListeners to the buttons is lost.

PR #1053, opened 2020-10-19
[FTAMS102020] Rich Alossi

alossix:


javascript/chronometer.js
```diff
@@ -1,26 +1,31 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
   startClick(callback) {
-    // ... your code goes here
```
LloydC, line 6: NIce work Rich! You just forgot to assign `setInterval` to your `intervalId` but besides that everything is correct
```
      this.currentTime += 1;
    }, 1000);

PR #1054, opened 2020-10-19
[FTAMS102020] Sophie Tichelaar

Sophietic:
iteration1 done

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,40 +1,58 @@
 class Chronometer {
   constructor() {
-    this.currentTime = 0;
-    this.intervalId = 0;
+   this.currentTime = 0;
+   this.intervalId = 0;  //waarom is deze 0?
   }
-  startClick(callback) {
-    setInterval(
-      () => this.currentTime++, 1000)
+  
+  //Verhoogt tijd elke sec met 1. gaat callback krijgen om tijd te printen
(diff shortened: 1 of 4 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,40 @@
...
   getMinutes() {
-    // ... your code goes here
+    // current time moet naar minuten
+    let minutes = this.currentTime / 60
+    return Math.floor(minutes)
   }
+
   getSeconds() {
-    // ... your code goes here
+    // minutes moeten naar seconden, dus *60
```
LloydC, line 19: Good work Sophie! You can make your solution even easier if you use a modulo operator `%` if you're unfamiliar with it, check out [this link](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Math))
```
return this.currentTime % 60;

File: javascript/chronometer.js [hunk 2]
```diff
@@ -1,26 +1,40 @@
...
+    return seconds;
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+  twoDigitsNumber(number) {
+    if (number < 10) {
+      return `0${number}`
+    } else { return number }
   }
+
```
LloydC, line 28: How to write the same solution using the ternary operator we introduced yesterday
```
return number < 10 ? `0${number}`: number;

PR #1055, opened 2020-10-19
[FTAMS102020] Emma Northridge

emmanorthridge:
Iterations complete

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -7,18 +7,15 @@ class Chronometer {
     this.intervalId = setInterval(() => {
       this.currentTime++;
     }, 1000);
-    }
-  
+  }
   getMinutes() {
-    return Math.floor(this.currentTime/60);
+    return Math.floor(this.currentTime / 60);
   }
   getSeconds() {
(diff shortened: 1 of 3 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,32 @@
...
   getSeconds() {
-    // ... your code goes here
+    return Math.floor(this.currentTime % 60);
   }
   twoDigitsNumber() {
-    // ... your code goes here
+    if (this.currentTime < 10) {
+      return ("0" + this.currentTime);
+    }
   }
```
LloydC, line 22: It's better practice if you pass "this.currentTIme" as an argument to be passed to the function `twoDigitsNumber(timeValue){...}`

You should also have an `else{ return timeValue}`. It can all be rewritten using the ternary operator 
```
twoDigitsNumber(timeValue) {
return timeValue < 10 ? "0" + timeValue : timeValue;
}

emmanorthridge:
Thanks @LloydC.

I have updated with your comments - they make sense. Will try do keep that in mind for future.

PR #1056, opened 2020-10-19
[WDFT-AMS-102020] - Maya Ausset

MayaAusset:


javascript/chronometer.js
```diff
@@ -1,26 +1,33 @@
...
-    // ... your code goes here
+    let minutes = Math.floor(this.currentTime / 60);
+    return minutes
   }
   getSeconds() {
-    // ... your code goes here
+    let minutes = Math.floor(this.currentTime / 60);
+    let seconds = this.currentTime - minutes * 60;
+    return seconds
   }
```
LloydC, line 17: Good work Maya! You can make your solution even easier if you use a modulo operator % if you're unfamiliar with it, check out [this link](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Math#Arithmetic_operators)
```
return this.currentTime % 60;

File: javascript/chronometer.js [hunk 2]
```diff
@@ -1,26 +1,33 @@
...
   twoDigitsNumber() {
-    // ... your code goes here
+  let myNumber = this.currentTime;
+  let formattedNumber = ("0" + myNumber)
+  return formattedNumber
+
   }
   stopClick() {
-    // ... your code goes here
+    clearInterval()
```
LloydC, line 25: In order to clear the Interval, you must pass the current intervalId as well `clearInterval(this.intervalId);`

javascript/chronometer.js
```diff
@@ -1,26 +1,33 @@
...
+    let minutes = Math.floor(this.currentTime / 60);
+    let seconds = this.currentTime - minutes * 60;
+    return seconds
   }
   twoDigitsNumber() {
-    // ... your code goes here
+  let myNumber = this.currentTime;
+  let formattedNumber = ("0" + myNumber)
+  return formattedNumber
+
```
LloydC, line 22: Your current solution only works when `this.currentTime < 10`. Here's a solution that also both scenarios
```
twoDigitsNumber(value) {
    if (value < 10) {return `0${value}`;}
    else{return `${value}`;}
  }

PR #1058, opened 2020-10-19
[FT-AMS-102020] Elia Chiarucci

eliachiarucci:


javascript/chronometer.js
```diff
@@ -1,26 +1,28 @@
...
   startClick(callback) {
-    // ... your code goes here
+    this.intervalId = setInterval(() => this.currentTime++, 1000);
   }
   getMinutes() {
-    // ... your code goes here
+    return Math.floor(this.currentTime/60);
   }
   getSeconds() {
-    // ... your code goes here
```
LloydC, line 12: I don't think that `Math.floor()` is necessary here, the remainder will always be a whole number so I don't think there's need to use it here

javascript/chronometer.js
```diff
@@ -1,26 +1,28 @@
...
+    return Math.floor(this.currentTime % 60);
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+  twoDigitsNumber(number) {
+    number = number.toString();
+    return number.length !== 2 ? "0" + number : number;
   }
```
LloydC, line 18: Nice, another way you can write it is by using template stings so you won't need line 16 
```
return number.length !== 2 ? `0${number}` : `${number}`;

PR #1059, opened 2020-10-19
[FTAMS102020] Nina van den Braak

ninsajala:


javascript/chronometer.js
```diff
@@ -1,26 +1,35 @@
...
-    // ... your code goes here
+    setInterval(() =>
+      this.currentTime++, 1000)
   }
   getMinutes() {
-    // ... your code goes here
+    return Math.floor(this.currentTime / 60)
   }
   getSeconds() {
-    // ... your code goes here
```
LloydC, line 12: Good work Nina! You can make your solution even easier if you use a modulo operator `%` if you're unfamiliar with it, check out [this link](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Math#Arithmetic_operators)
```
return this.currentTime % 60;

File: javascript/chronometer.js [hunk 2]
```diff
@@ -1,26 +1,35 @@
...
+    return (this.currentTime - Math.floor(this.getMinutes() * 60))
   }
-  twoDigitsNumber() {
-    // ... your code goes here
```
LloydC, line 15: Nice, I think you don't need line 17 if you make use of template strings on line 21 `return `${number}` ;`

Also, if you use the ternary operator that we talked about yesterday then you can get a one-line solution (I'll use a different condition here as well)
```
return number < 10 ?  `0${number}` : `${number}`;

File: javascript/chronometer.js [hunk 3]
```diff
@@ -1,26 +1,35 @@
...
   stopClick() {
-    // ... your code goes here
+    clearInterval(this.IntervalId)
   }
   resetClick() {
-    // ... your code goes here
+    this.currentTime = 0
   }
   splitClick() {
-    // ... your code goes here
```
LloydC, line 24: You're missing the parantheses to call `getMinutes` and `getSeconds`, without them the functions will get accessed but they won't execute.

If you change that `splitClick` will work

javascript/chronometer.js
ninsajala: I tried to implement this but it's not working so well now with the milliseconds. I guess this.currentTime%6000 is not the answer, right?

javascript/chronometer.js
ninsajala: Of course, thanks!

PR #1060, opened 2020-10-19
[FTAMS102020] Irene Asensio

ihatepineapple:
1st Iteration complete

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,74 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    // declare variables currentTime and intervalId
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
   startClick(callback) {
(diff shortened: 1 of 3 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,74 @@
...
   resetClick() {
-    // ... your code goes here
+    this.currentTime = 0
   }
   splitClick() {
-    // ... your code goes here
+   
+    let min = this.twoDigitsNumber(this.getMinutes(this.currentTime))
+    let sec = this.twoDigitsNumber(this.getSeconds(this.currentTime))
+
```
LloydC, line 68: You don't need to pass `this.currentTime` on lines 66-67, if you check how these functions have been defined (without parameters) then you should see it's not necessary

ihatepineapple: thanks! but without it, jasmine was not testing ok

PR #1061, opened 2020-10-19
[FTAMS102020] Roy-Arne Brussel

Roybrussel:


javascript/chronometer.js
```diff
@@ -1,26 +1,35 @@
...
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
+
   startClick(callback) {
-    // ... your code goes here
+    setInterval(
+      () => this.currentTime++, 1000);
```
wax-bear, line 9: Great stuff implementing the setInterval method!
I would suggest two things:
- implement the call of the callback (conditionally) as mentioned in the iteration
- declare the function (the first argument of the setInterval method) outside of setInterval. Then, use it as the first argument. This makes the setInterval more readable.

javascript/chronometer.js
```diff
@@ -1,26 +1,35 @@
...
   }
   getMinutes() {
-    // ... your code goes here
+    return Math.floor(this.currentTime / 60)
   }
   getSeconds() {
-    // ... your code goes here
+    return (this.currentTime - Math.floor(this.getMinutes() * 60))
```
wax-bear, line 15: Could the modulo % operator be used here to minimize the code?

javascript/chronometer.js
```diff
@@ -1,26 +1,35 @@
...
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+  twoDigitsNumber(number) {
+    if (number < 10) {
```
wax-bear, line 18: This function should return a string. The function name as mentioned in the iteration is therefore confusing. It would be better named something like "getTwoDigitsString".

The code could be minimized as follows:
return (number < 10 ? '0' : '') + number;

PR #1064, opened 2020-10-20
[FT-AMS-102020] Emad Ghnim

Emad-doz:


javascript/chronometer.js
```diff
@@ -1,26 +1,38 @@
...
-    // ... your code goes here
+
+  getMilliseconds(){
+    return this.currentTimeMil.toString().slice(-2)
   }
+
+  twoDigitsNumber(number) {
+    return number.toString().length === 2 ? number.toString() : '0' + number.toString();
+    } 
+
```
LloydC, line 26: Good work Emad!

To turn a number into a string, you can use template strings. You can also use a different condition
```
twoDigitsNumber(number) {
    return number > 10? `${number}` : `0${number}`;
    }

File: javascript/chronometer.js [hunk 2]
```diff
@@ -1,26 +1,38 @@
...
+    clearInterval(this.intervalId);
+    clearInterval(this.intervalIdMil);
   }
   resetClick() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.currentTimeMil = 0;
   }
   splitClick() {
-    // ... your code goes here
```
LloydC, line 24: You can use template strings as well here
```
return `${this.twoDigitsNumber(this.getMinutes())}:${this.twoDigitsNumber(this.getSeconds())}:${this.twoDigitsNumber(this.getMilliseconds())}`;

PR #1068, opened 2020-10-20
[FTAMS102020] CeydaUlubas-lab-javascript-chronometer

ceydaulubas:
completed only Iteration1

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,35 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
-  startClick(callback) {
-    // ... your code goes here
+
+  startClick(){
+    this.intervalId= setInterval(() => { 
+      this.currentTime ++;
+    },1000);
   }
+
   getMinutes() {
-    // ... your code goes here
+    return Math.floor(this.currentTime / 60);
   }
   getSeconds() {
-    // ... your code goes here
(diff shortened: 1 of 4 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,35 @@
...
+    return Math.floor(this.currentTime / 60);
   }
   getSeconds() {
-    // ... your code goes here
+    return this.currentTime % 60;
   }
   twoDigitsNumber() {
-    // ... your code goes here
+    return this.currentTime.toString( ).padStart(2, '0');
   }
```
LloydC, line 21: It would be better if you made your function twoDigitsNumber to accept arguments, that way it will work for a number of use cases.

Also to make sure that we add a 0 when the value is between 0 and 10, we can write a condition
```
twoDigitsNumber(value) {
    if (value < 10) return `0${value}`;
    return `${value}`;
  }

File: javascript/chronometer.js [hunk 2]
```diff
@@ -1,26 +1,35 @@
...
   stopClick() {
-    // ... your code goes here
+    clearInterval(this.intervalId)
   }
   resetClick() {
-    // ... your code goes here
+    this.currentTime=0
   }
   splitClick() {
-    // ... your code goes here
```
LloydC, line 24: Nice, you can also write this using template strings
```
return `${this.twoDigitsNumber(this.getMinutes())} :${this.twoDigitsNumber(this.getSeconds())}`;

PR #1079, opened 2020-10-21
[FTAMS1020202]Mahalakshmi

mahavinay:
"Completed till Iteration4 except clearing the child elements and after stop reset is not done"

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,58 @@
 class Chronometer {
   constructor() {
     // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = 0
   }
   startClick(callback) {
+    this.intervalId = setInterval(
+      () => {
+        this.currentTime++;
+            if (callback) {
+                callback()
+            }
+    }, 1000)
+    //console.log("this.startClick")
+}
+
+    
     // ... your code goes here
-  }
+    
+  
+  //setInterval = (startClick() => {),1000}
   getMinutes() {
     // ... your code goes here
+    const minutes = Math.floor(this.currentTime / 60);
+    return minutes
+
   }
   getSeconds() {
-    // ... your code goes here
+    return Math.floor(this.currentTime % 60);
+ 
   }
-  twoDigitsNumber() {
+  twoDigitsNumber(number) {
     // ... your code goes here
+    
+  
+    return number > 9 ? String(number) : `0`+ number;
+    
+    
   }
   stopClick() {
     // ... your code goes here
+    clearInterval(this.intervalId)
   }
   resetClick() {
     // ... your code goes here
+    this.currentTime=0;
   }
   splitClick() {
     // ... your code goes here
-  }
+    let min= this.twoDigitsNumber(this.getMinutes());
+    let sec=this.twoDigitsNumber(this.getSeconds());
+    return `${min}:${sec}`;
+   
+    } 
 }
+
+//Iteration 2: DOM Manipulation
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -40,7 +40,7 @@ class Chronometer {
   }
   stopClick() {
     // ... your code goes here
-    clearInterval(this.intervalId)
+    clearInterval(this.intervalId);
   }
   resetClick() {
     // ... your code goes here
--- a/javascript/index.js
+++ b/javascript/index.js
@@ -15,14 +15,34 @@ let splits = document.getElementById('splits');
 
 function printTime() {
   // ... your code goes here
+
+  // WHat you want in the end is to call both printMinutes
+  // and printSeconds in here
+
+  // printMinutes();
+  let fetchedSec= printSeconds();
+  secDec.innerHTML=fetchedSec[0];
+  secUni.innerHTML=fetchedSec[1];
+  let fetchedMin=printMinutes();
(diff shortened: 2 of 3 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,58 @@
...
+  //setInterval = (startClick() => {),1000}
   getMinutes() {
     // ... your code goes here
+    const minutes = Math.floor(this.currentTime / 60);
+    return minutes
+
   }
   getSeconds() {
-    // ... your code goes here
+    return Math.floor(this.currentTime % 60);
```
LloydC, line 30: There's no need to use `Math.floor()` here as the remainder will always be a whole integer
```
return this.currentTime % 60;

File: javascript/chronometer.js [hunk 2]
```diff
@@ -1,26 +1,58 @@
...
+ 
   }
-  twoDigitsNumber() {
+  twoDigitsNumber(number) {
     // ... your code goes here
+    
+  
+    return number > 9 ? String(number) : `0`+ number;
+
```
LloydC, line 38: Nice, here's another way to write your return statement using a different condition and template strings
```
return number < 10 ? `0${number}` : `${number}`

File: javascript/index.js [hunk 3]
```diff
@@ -32,14 +52,26 @@ function printMilliseconds() {
...
+  while (splits.hasChildNodes()) {  
+    splits.removeChild(splits.firstChild);
+  }
+  
+  
 }
 
 function setStopBtn() {
   // ... your code goes here
+  chronometer.stopClick;
```
LloydC, line 73: On line 104, you're already firing `chronometer.stopClick()` so you don't need this line here.

But you could have used function setStopBtn for DOM manipulation
```
function setStopBtn() {
  btnLeft.className = 'btn stop';
  btnLeft.textContent = 'STOP';
}

File: javascript/index.js [hunk 4]
```diff
@@ -48,18 +80,63 @@ function setSplitBtn() {
...
+
+ document.getElementById('btnLeft').addEventListener("click", () => {
+  const btnLeftClassName = document.getElementById('btnLeft').className;
+  if(btnLeftClassName==="btn start"){
+   chronometer.startClick(printTime);
+  btnLeft.innerHTML="STOP";
+  btnLeft.className="btn stop"
+  btnRight.className="btn split"
+  btnRight.innerHTML="SPLIT"
+  chronometer.stopClick()
```
LloydC, line 100: You should remove line 100

javascript/index.js
```diff
@@ -48,18 +80,63 @@ function setSplitBtn() {
...
+  }
+
+ if(btnLeftClassName==="btn stop"){
+  chronometer.stopClick()
+  btnLeft.innerHTML="STOP";
```
LloydC, line 105: That's incorrect, after clicking on "STOP" you want the button to turn back to "START"
```
btnLeft.className = 'btn start';
btnLeft.textContent = 'START';

File: javascript/index.js [hunk 6]
```diff
@@ -48,18 +80,63 @@ function setSplitBtn() {
...
+  btnRight.className="btn split";
+  btnLeft.innerHTML="STOP"
+   chronometer.resetClick();
+    clearSplits();
+  
+   }
+
+  if(btnRighttClassName==="btn split"){
+    printSplit();
+    btnRight.innerHTML="RESET";
```
LloydC, line 133: After you click on "SPLIT", the only thing that should happen is executing `printSplit()` so the current time gets added to the list.

You don't need line 133-136, that behaviour is taken care of already with the btnLeft event listener.

javascript/index.js
```diff
@@ -48,18 +80,63 @@ function setSplitBtn() {
...
+  if(btnRighttClassName==="btn reset"){
+  btnRight.innerHTML="SPLIT";
+  btnLeft.className="btn stop";
+  btnRight.className="btn split";
+  btnLeft.innerHTML="STOP"
+   chronometer.resetClick();
+    clearSplits();
+  
+   }
+
```
LloydC, line 130: Since you have only 2 use cases, you could have used `else{}` instead of another `if` condition.

javascript/index.js
```diff
@@ -32,14 +52,26 @@ function printMilliseconds() {
...
+  }
+  
+  
 }
 
 function setStopBtn() {
   // ... your code goes here
+  chronometer.stopClick;
+
 }
```
LloydC, line 76: Similarly, functions setSplitBtn,setStartBtn and setResetBtn can also be used for handling DOM manipulation
```
function setSplitBtn() {
  btnRight.className = 'btn split';
  btnRight.textContent = 'SPLIT';
}

function setStartBtn() {
  btnLeft.className = 'btn start';
  btnLeft.textContent = 'START';
}

function setResetBtn() {
  btnRight.className = 'btn reset';
  btnRight.textContent = 'RESET';
}

File: javascript/index.js [hunk 9]
```diff
@@ -32,14 +52,26 @@ function printMilliseconds() {
...
   // ... your code goes here
+  var childElement = document.createElement("LI");
+  childElement.innerHTML = chronometer.splitClick()
+  splits.appendChild(childElement);
+
 }
 
 function clearSplits() {
   // ... your code goes here
+
```
LloydC, line 63: You don't necessarily have to check if there are nodes. Also another way you can reset the splits is by using `splits.innerHTML = ' '`

PR #1080, opened 2020-10-21
[FTAMS102020]  Sabela Pena Garcia

defSabe:
most of it done, but still issues with btn actions

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,43 @@
 class Chronometer {
+
   constructor() {
-    // ... your code goes here
-  }
+    this.currentTime = 0
+    this.intervalId = 0
+
+  };
+
   startClick(callback) {
-    // ... your code goes here
+    console.log("startClick")
+    this.intervalId = setInterval(() => { // arrow function is the callback
+      this.currentTime++;
+      if (callback) callback();
+    }, 1000);
+
   }
+
   getMinutes() {
-    // ... your code goes here
+    return Math.floor(this.currentTime / 60); //61 seconds, 122 seconds
+
   }
   getSeconds() {
-    // ... your code goes here
+    return (this.currentTime % 60);
+
   }
-  twoDigitsNumber() {
-    // ... your code goes here
+  twoDigitsNumber(num) {
+    return num.toString().padStart(2, '0');
+    // return this.currentTime.toString().padStart(2, '0');
   }
   stopClick() {
-    // ... your code goes here
+    clearInterval(this.intervalId);
   }
   resetClick() {
-    // ... your code goes here
+    this.currentTime = 0;
   }
   splitClick() {
-    // ... your code goes here
+    let min = this.twoDigitsNumber(this.getMinutes());
+    let sec = this.twoDigitsNumber(this.getSeconds());
+    console.log("min and sec " + min + sec);
+    return `${min}:${sec}`;
+    //    return `${this.twoDigitsNumber(this.getMinutes)}\:${this.twoDigitsNumber(this.getSeconds)}`;
   }
 }
--- a/javascript/index.js
+++ b/javascript/index.js
@@ -1,29 +1,56 @@
 const chronometer = new Chronometer();
-
 // get the buttons:
+
 const btnLeft = document.getElementById('btnLeft');
-const btnRight = document.getElementById('btnRight');
+btnLeft.onclick = function () {
+  if (btnLeft.className === "btn start") {
+    chronometer.startClick(printTime);
+    // printTime as a callback
+    // two functions printMinutes and printSeconds
+  } else if (btnLeft.className === "btn stop") {
+    chronometer.stopClick()
+  }
+};
+const btnRight = document.getElementById("btnRight");
+btnRight.onclick = function () {
+  if (btnRight.className === "btn reset") {
+    chronometer.resetClick()
+  } else if (btnRight.className === "btn split") {
+    chronometer.splitClick()
+  }
+}
+
+// const splitTime = document.getElementById("splits")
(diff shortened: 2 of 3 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,43 @@
...
+    return Math.floor(this.currentTime / 60); //61 seconds, 122 seconds
+
   }
   getSeconds() {
-    // ... your code goes here
+    return (this.currentTime % 60);
+
   }
-  twoDigitsNumber() {
-    // ... your code goes here
```
LloydC, line 15: Here is another working solution for twoDigitsNumber
```
twoDigitsNumber(num) {
if(num <10){
return '0' + number;
}
else {
  return number.toString()
}
}

File: javascript/chronometer.js [hunk 2]
```diff
@@ -1,26 +1,43 @@
...
+    this.currentTime = 0;
   }
   splitClick() {
-    // ... your code goes here
+    let min = this.twoDigitsNumber(this.getMinutes());
+    let sec = this.twoDigitsNumber(this.getSeconds());
+    console.log("min and sec " + min + sec);
+    return `${min}:${sec}`;
+    //    return `${this.twoDigitsNumber(this.getMinutes)}\:${this.twoDigitsNumber(this.getSeconds)}`;
```
LloydC, line 41: This is fine, I just want to point out that you should remove `console.logs` and commented out code once you're finished.

It will make your code easier to read and minimal

javascript/index.js
```diff
@@ -1,30 +1,57 @@
 const chronometer = new Chronometer();
-
 // get the buttons:
+
 const btnLeft = document.getElementById('btnLeft');
-const btnRight = document.getElementById('btnRight');
+btnLeft.onclick = function () {
+  if (btnLeft.className === "btn start") {
+    chronometer.startClick(printTime);
+    // printTime as a callback
```
LloydC, line 8: You also want the buttons to change their colour and content after you click on start, so you need to add some code for DOM manipulation
```
btnLeft.className = 'btn stop';
btnLeft.textContent = 'STOP';
btnRight.className = 'btn split';
btnRight.textContent = 'SPLIT';

File: javascript/index.js [hunk 4]
```diff
@@ -1,30 +1,57 @@
...
+    // two functions printMinutes and printSeconds
```
LloydC, line 9: Since you have 2 uses cases (START and STOP), you can use an `else{...}` here

javascript/index.js
```diff
@@ -1,30 +1,57 @@
...
+  } else if (btnLeft.className === "btn stop") {
+    chronometer.stopClick()
```
LloydC, line 11: Just like before, we want our buttons to change accordingly using some DOM manipulation
```
btnLeft.className = 'btn start';
btnLeft.textContent = 'START';
btnRight.className = 'btn reset';
btnRight.textContent = 'RESET';

File: javascript/index.js [hunk 6]
```diff
@@ -1,30 +1,57 @@
...
+  minDec.innerHTML = chronometer.twoDigitsNumber(chronometer.getMinutes())[0];
 }
 
 function printSeconds() {
-  // ... your code goes here
+  secUni.innerHTML = chronometer.twoDigitsNumber(chronometer.getSeconds())[1];
+  secDec.innerHTML = chronometer.twoDigitsNumber(chronometer.getSeconds())[0];
 }
 
+
```
LloydC, line 53: For printing the splits time,  what you want to do is get access to the ordered list element by pointing to its unique identifier "splits" (it's been done on line 35) then you're going to add a <li> using dome DOM manipulation again
```
function printSplit() {
  let li = document.createElement('li'); 
  li.innerHTML = chronometer.splitClick();
  splits.appendChild(li);
}

File: javascript/index.js [hunk 7]
```diff
@@ -1,30 +1,57 @@
...
+  } else if (btnLeft.className === "btn stop") {
+    chronometer.stopClick()
+  }
+};
+const btnRight = document.getElementById("btnRight");
+btnRight.onclick = function () {
+  if (btnRight.className === "btn reset") {
+    chronometer.resetClick()
+  } else if (btnRight.className === "btn split") {
+    chronometer.splitClick()
```
LloydC, line 19: I added a function `printSplit` on line 53 which already makes use of your `chronometer.splitClick()` so you can replace it with `printSplit()` and it should work

javascript/index.js
```diff
@@ -1,30 +1,57 @@
...
 }
 
 function printSeconds() {
-  // ... your code goes here
+  secUni.innerHTML = chronometer.twoDigitsNumber(chronometer.getSeconds())[1];
+  secDec.innerHTML = chronometer.twoDigitsNumber(chronometer.getSeconds())[0];
 }
 
+
+
```
LloydC, line 54: For clearing the splits, what you want is basically to access the same element with a "splits" identifier and remove any list item by doing some DOM manipulation again
```
function clearSplits() {
  splits.innerHTML = '';
}

File: javascript/index.js [hunk 9]
```diff
@@ -1,30 +1,57 @@
...
+    // printTime as a callback
+    // two functions printMinutes and printSeconds
+  } else if (btnLeft.className === "btn stop") {
+    chronometer.stopClick()
+  }
+};
+const btnRight = document.getElementById("btnRight");
+btnRight.onclick = function () {
+  if (btnRight.className === "btn reset") {
+    chronometer.resetClick()
```
LloydC, line 17: I added a function to clear the splits on line 54, it should be used to complement the reset action. You'll then also need to display the reset time again by calling `printTime`
```
if (btnRight.classList.contains('reset')) {
    chronometer.resetClick();
    clearSplits();
    printTime();
}

PR #1081, opened 2020-10-22
[FTAMS102020]issaGruber

GruberI:


javascript/chronometer.js
```diff
@@ -1,26 +1,52 @@
...
+    return this.currentTime % 60
+  }
+
+  twoDigitsNumber(number) {
+    console.log(number);
+    if (number < 10) {
+      return `0${number}`
+    } else {
+      return `${number}`;
+    }
```
reMRKableDev, line 31: **ES6 tip**: A nice thing you could do here is to make use of the ternary operator. It's especially helpful when you have an ```if ...else``` statement that only does a singular thing within each code block. Check out the suggestion below and the video I've linked to get a better understanding of what I mean

```suggestion
    return number < 10 ? `0${number}` : `${number}`;
```

Find video here --> https://www.youtube.com/watch?v=70GmtV-5Ggo&ab_channel=WebDevSimplified

javascript/index.js
```diff
@@ -14,52 +14,84 @@ let milUni = document.getElementById('milUni');
...
 }
 
 function printSplit() {
-  // ... your code goes here
+  let fetchSplit = chronometer.splitClick();
+  let newSplitsLi = document.createElement("li")
+   splits.appendChild(newSplitsLi).innerHTML = fetchSplit;
 }
 
 function clearSplits() {
```
reMRKableDev, line 46: This function needs to be done so that the reset button can be used on your chronometer 😄

reMRKableDev (changes requested):
Nice work @GruberI 🚀 

Your chronometer.js solutions are good. I've given you a small pro tip that you can use. Chekc it out. 

The only solution missing for iteration 4 which will make your Reset button work. You will need to provide a solution for the ```clearSplits()``` function in ```index.js```

PR #1082, opened 2020-10-23
[FTAMS102020] AlejandroPliakasBenitez

apliakas:


javascript/chronometer.js
```diff
@@ -1,26 +1,51 @@
...
+  }
+
   getMinutes() {
-    // ... your code goes here
+    return Math.floor(this.currentTime / 60);
   }
+
   getSeconds() {
-    // ... your code goes here
+    return this.currentTime - (Math.floor(this.currentTime / 60) * 60);
```
LloydC, line 29: You can get a simpler solution by using the [modulo operator](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Math#Arithmetic_operators)
```
return this.currentTime % 60;

PR #1083, opened 2020-10-23
[WDFT-AMS-102020]-Alex Kantsadis

teof95:
Split does not work yet

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,26 +1,41 @@
 class Chronometer {
-  constructor() {
-    // ... your code goes here
-  }
-  startClick(callback) {
-    // ... your code goes here
-  }
-  getMinutes() {
-    // ... your code goes here
-  }
-  getSeconds() {
-    // ... your code goes here
-  }
-  twoDigitsNumber() {
-    // ... your code goes here
-  }
-  stopClick() {
-    // ... your code goes here
-  }
-  resetClick() {
-    // ... your code goes here
-  }
-  splitClick() {
-    // ... your code goes here
-  }
+    constructor() {
+        this.currentTime = 0;
+        this.intervalId = 0;
+    }
+    startClick(callback) {
+        this.intervalId = setInterval(() => {
+            this.currentTime += 1;
+            if (callback) callback();
+        }, 1000);
+    }
+
+    getMinutes() {
+        let minutes = Math.floor(this.currentTime / 60)
+        return minutes
+    }
+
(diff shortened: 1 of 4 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,41 @@
...
+
+    getMinutes() {
+        let minutes = Math.floor(this.currentTime / 60)
+        return minutes
+    }
+
+    getSeconds() {
+        if (!this.currentTime)
+            return 0;
+        return this.currentTime - this.getMinutes() * 60;
```
LloydC, line 21: You can get a simpler solution by using the [modulo operator](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Math#Arithmetic_operators)
```
return this.currentTime % 60;

File: javascript/index.js [hunk 2]
```diff
@@ -14,52 +14,85 @@ let milUni = document.getElementById('milUni');
...
+// function printMilliseconds() {
+//     const millis = chronometer.twoDigitsNumber(chronometer.getMilliseconds());
+//     milDec.innerText = millis[0];
+//     milUni.innerText = millis[1];
+// }
 
 function printSplit() {
-  // ... your code goes here
+    let list = document.createElement('li');
+    list.innerHTML = chronometer.setTime();
```
LloydC, line 43: You used the wrong function call on line 43
```
list.innerHTML = chronometer.splitClick();

File: javascript/index.js [hunk 3]
```diff
@@ -14,52 +14,85 @@ let milUni = document.getElementById('milUni');
...
+    parent.appendChild(list);
 }
 
 function clearSplits() {
-  // ... your code goes here
+    document.getElementById("splits").innerHTML = "";
 }
 
 function setStopBtn() {
```
LloydC, line 51: The logic defined line 52-54 is for setting the "start" button, therefor you should either rename your function or switch the logic with function setStartBtn

javascript/index.js
```diff
@@ -14,52 +14,85 @@ let milUni = document.getElementById('milUni');
...
 
 function setSplitBtn() {
-  // ... your code goes here
+    btnRight.classList.remove('reset');
+    btnRight.classList.add('split');
+    btnRight.innerHTML = 'SPLIT';
 }
 
 function setStartBtn() {
-  // ... your code goes here
```
LloydC, line 50: Same comment as above, either rename this function or switch its logic with function setStopBtn

PR #1084, opened 2020-10-23
[FTAMS102020] Ahmet Demirlicakmak

codebrokerad:
timer is working butsplit times does not print also have some other issues that I could not fix

### DIFF
--- a/javascript/index.js
+++ b/javascript/index.js
@@ -14,15 +14,21 @@ let milUni = document.getElementById('milUni');
 let splits = document.getElementById('splits');
 
 function printTime() {
-  // ... your code goes here
-}
 
+  printMinutes();
+  printSeconds();
+
+}
 function printMinutes() {
-  // ... your code goes here
+  const mins = chronometer.twoDigitsNumber(chronometer.getMinutes());
+    minDec.innerText = mins[0];
+    minUni.innerHTML = mins[1];
 }
 
 function printSeconds() {
-  // ... your code goes here
+  const secs = chronometer.twoDigitsNumber(chronometer.getSeconds());
+  secDec.innerText = secs[0];
+  secUni.innerHTML = secs[1];
 }
 
 // ==> BONUS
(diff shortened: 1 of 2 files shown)


javascript/index.js
```diff
@@ -14,52 +14,91 @@ let milUni = document.getElementById('milUni');
...
 // ==> BONUS
 function printMilliseconds() {
   // ... your code goes here
 }
 
-function printSplit() {
-  // ... your code goes here
-}
+  function printSplit() {    
+    let splitsCount = document.getElementById('splits')
```
LloydC, line 40: You already have `let splits` on line 14 so this is unnecessary, you can replace line 43 with `splitsCount.appendChild(newSplit)`

javascript/index.js
```diff
@@ -14,52 +14,91 @@ let milUni = document.getElementById('milUni');
...
+    splitsCount.appendChild(newSplit)}
+
 
 function clearSplits() {
-  // ... your code goes here
+  document.getElementById("splits").innerHTML = "";
 }
 
 function setStopBtn() {
-  // ... your code goes here
```
LloydC, line 42: The logic defined line 51-53 is for setting the "start" button, therefor you should either rename your function or switch the logic with function setStartBtn

javascript/index.js
```diff
@@ -14,52 +14,91 @@ let milUni = document.getElementById('milUni');
...
 
 function setSplitBtn() {
-  // ... your code goes here
+    btnRight.classList.remove('reset');
+    btnRight.classList.add('split');
+    btnRight.innerHTML = 'SPLIT';
 }
 
 function setStartBtn() {
-  // ... your code goes here
```
LloydC, line 50: Same comment as above, either rename this function or switch its logic with function setStopBtn

javascript/index.js
```diff
@@ -14,52 +14,91 @@ let milUni = document.getElementById('milUni');
...
+    btnRight.classList.remove("split");
+    btnRight.classList.add("reset");
+    btnRight.innerHTML = "RESET";
 }
 
 // Start/Stop Button
 btnLeft.addEventListener('click', () => {
-  // ... your code goes here
+  if (btnLeft.classList.contains("start")){
+    btnLeft.className = "btn stop";
```
LloydC, line 77: You can replace line 77-80
```
setStopBtn();
setSplitBtn();

File: javascript/index.js [hunk 5]
```diff
@@ -14,52 +14,91 @@ let milUni = document.getElementById('milUni');
...
+    btnLeft.textContent = "STOP";
+    btnRight.className = "btn split";
+    btnRight.textContent = "SPLIT";
+    chronometer.startClick();
+    let printIntervalId = setInterval (()=>{
+    printTime();
+    }, 1);
+  } else if (btnLeft.classList.contains("stop")){
```
LloydC, line 85: Since you only have two use cases, you can also use the `else` statement.

Line 86-89 can be replaced
```
setStartBtn();
setResetBtn();

File: javascript/index.js [hunk 6]
```diff
@@ -14,52 +14,91 @@ let milUni = document.getElementById('milUni');
...
+    btnRight.className = "btn reset";
+    btnRight.textContent = "RESET";
+    chronometer.stopClick();
+  }
 });
 
 // Reset/Split Button
 btnRight.addEventListener('click', () => {
-  // ... your code goes here
+  if(btnRight.classList.contains("reset")){
```
LloydC, line 96: You need to call `printTime()` for the reset to display "00:00"

javascript/index.js
```diff
@@ -14,52 +14,91 @@ let milUni = document.getElementById('milUni');
...
+    chronometer.resetClick();
+    clearSplits();
+  } else if (btnRight.classList.contains("split")){
```
LloydC, line 99: Again you can use an `else` statement here.

Also you don't need line 100-101 (you're calling `chronometer.splitClick()` within `printSplit()` already)

PR #1167, opened 2020-11-18
[PAR-WDFT-911] Elisa CHEN 

elisachen86:


elisachen86:
Please refer to the second pull request (lab completed without bonus) for the completed bonus. Thanks!

elisachen86:
version with the bonus completed here. 
FYI, i think there are some jasmine tests that need to be changed for the bonus: 
1 / startClick
- should increment by 1 the currentTime property on every 1 second interval (no longer, with bonus should increment every 100th of a second)
- after 3 seconds, currentTime should be 3 (no, because see above)
2/ getMinutes
- should return the currentTime in minutes (the input and expected values are still based on a seconds interval)
- should return the currentTime in minutes even for laaaarge numbers (same as above)
3/ get seconds
- should return the seconds of the currentTime (same comment as above, input/expected values based on seconds interval)
- should return the seconds portion of the currentTime that remains after removing the minutes

PR #1314, opened 2021-04-01
Problem with milliseconds

AndyMcGN:
Still not worked out how to make chronometer count at the correct speed.


AndyMcGN:
Problem with ms sorted, will send a new pull request now!

PR #1317, opened 2021-04-01
[FTAMS032021] Igor Godinho Figueiredo

Igor-GF:
I was not able to print the numbers on the chronometer display.

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,8 +1,9 @@
 class Chronometer {
   constructor() {
-    this.currentTime = 152;
+    this.currentTime = 0;
     this.intervalId = 0;
   }
+  
   startClick(callback) {
     this.intervalId = setInterval(() => {
      this.currentTime++;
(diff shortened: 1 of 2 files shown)


javascript/chronometer.js
```diff
@@ -1,26 +1,50 @@
...
+    this.currentTime = 0;
+    this.intervalId = 0;
   }
+  
   startClick(callback) {
-    // ... your code goes here
-  }
+    this.intervalId = setInterval(() => {
+     this.currentTime++;
+    //  console.log(this.currentTime);
```
reMRKableDev, line 10: You can delete this `console.log()`

reMRKableDev (commented):
Nice work @Igor-GF 🚀 

So the tests are green which is good. On the `index.html` the timer starts, changing the text to STOP and vice-versa. Additionally, I am able to split the time and see the split time on the list next to the timer. 

The only thing right now that needs to be done is showing the time changes on the timer. You need to make use of the `printMinutes` and `printSeconds` methods. You've started on something with both which is going towards the right way. Once you've defined the functionality of these methods, call them inside of the `printTime` method. 

Subsequently, you can call `printTime` upon click the START button

PR #1378, opened 2021-06-09
Full Timer Lab

BartSpangenberg:
Hi TA's!

Lab was nice. I think the demo timer has a bug in it. It runs it for every ms but also displays the milliseconds (on the clock and in the  interval). This is kinda weird because leaves out the hundreds of milliseconds, therefore creating an inaccurate time.

It goes from 0-99 (0-99) then from 0-99 (100-199 ms) etc. this happens 10 times than it goes to 1 second.

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -44,7 +44,8 @@ class Chronometer {
     // return time formated as mm:ss
     let minutes =  this.computeTwoDigitNumber(this.getMinutes());
     let seconds = this.computeTwoDigitNumber(this.getSeconds());
-    return `${minutes}:${seconds}`;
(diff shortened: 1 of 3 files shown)


TA-Remote:
Hey Bart! You did awesome! No words 💯 This was a pretty hard lab and you've been able to overcome it very well! 
The truth is that I can't find mistakes or much to improve but I will try to give you some advice anyway:

- I like a lot how you solved the two digit number function. very smart solution. 👍 
- Being picky i would tell you to try to get rid of the useless comments before submitting. 
- Also, as you are doing great and I see you get how DOM works and your code is clean, I would encourage you to put some simple comments in your code explaining how things work (but keep them short and clear). 

Anyway Bart, you did awesome! Keep the hard work 🎖️ 👍 🥇

PR #1380, opened 2021-06-09
Mariana WDFT-RMT-052021

marileitune:


TA-Remote:
Hey Mariana! You did great! 💯 This was a pretty hard lab and you've been able to overcome it very well! I think you should have tried the bonus!
I will try to give you some advice anyway:

- Very nice to use the charAt method on the print functions. Very nice. 👍 
- Being picky i would tell you to try to get rid of the useless comments before submitting. 
- I like your code is really clean and readable. That's perfect :)
- Also, as you are doing great and I see you get how DOM works and your code is clean, I would encourage you to put some simple comments in your code explaining how things work (but keep them short and clear). 

Anyway Mariana!, you did great! Keep the hard work 🎖️ 👍 🥇

PR #1381, opened 2021-06-09
George WDFT-RMT-MAY21

gdsbrooks:
Actually not horrible. will implement ms timer this weekend.

### DIFF
--- a/index.html
+++ b/index.html
@@ -37,7 +37,7 @@
       <div class="hole"></div>
     </div>
 
-    <script src="javascript/chronometer.js"></script>
+    <script src="/javascript/chronometer.js"></script>
     <script src="javascript/index.js"></script>
   </body>
 </html>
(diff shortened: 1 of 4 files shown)


TA-Remote:
Hey George! You did great! 💯 This was a pretty hard lab and you've been able to overcome it very well! 
The truth is that I can't find mistakes or much to improve but I will try to give you some advice anyway:
- I saw you used padStart method for the two digits number function, very smart. I like it 👍 
- Being picky i would tell you to try to be consistent with semicolons. Either on all lines or none at all. I saw you missed a few ones. 
- Also, as you are doing great and I see you get how DOM works and your code is clean, I would encourage you to put some simple comments in your code. 
Anyway George, you did awesome! Keep the hard work! 🎖️

gdsbrooks:
Thank you!

Yesterday I had to travel down to Lisbon so did the code on the train/bus/as my boo was cooking dinner. 

If I’d had the entire evening I would have refactored and beautified.

Part of me likes semicolons part of me doesn’t. It seems inconsistent and unclear in some of our labs. My sense is I *should* do it, but don’t *need*.

Thank you for the feedback!

PR #1385, opened 2021-06-09
[WDFT-RMT-MAY21] Conchita

CodeChita:
Hi TA! 

Got it working! I understand in pseudocode what to do for the bonus. I just don't got the energy to do it :)
**Question of the day: what is your go music you play while coding? Mine is LoFi beats** 

Have a great day!!

### DIFF
--- a/index.html
+++ b/index.html
@@ -10,7 +10,8 @@
   <body>
     <section id="splits-container">
       <h3>Splits</h3>
-      <ol id="splits"></ol>
+      <ol id="splits">
+      </ol>
     </section>
 
     <div class="leash leash-top"></div>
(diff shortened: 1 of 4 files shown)


TA-Remote:
Hey Conchita! 
Nice job on the lab. The chrono works which is cool however the way you did it prevent you from passing the jest tests. That is fine though as it works. I really recommend you redo the lab with the solution to see how you could have passed the tests :) 
I would also add that you could give more meaningful names to your variables. For example in the computeTwoDigitNumber function you have 2 let = str . Get into the habit of giving proper names to your variables as it will really help when you work with other people :) 
And about your question I usually don't listen to music while I code except when I do the CSS part and I put an electro vibe. 
Have a nice day!

PR #1386, opened 2021-06-09
Joanne Seiler WDFT-RMT-MAY21

Joanneseiler:


TA-Remote:
Hey Joanne, 
Very well done on the lab. Everything works!
I just wanted to tell you that regarding the logic behind the buttons (changing the class and names), it is better to put that logic inside the appropriate button functions and keep your addEventListener clean ;) 
Other than that very well done!

PR #1387, opened 2021-06-09
Halil Ibrahim WDFT-RMTMAY21

halokaya67:


TA-Remote:
Hey Halil!

Good work on the chronometer.js file, your logic is good and your code is clean. Also for the index.js, you didn't go with the callbacks which can help your code being reusable. Your logic is not wrong but you forgot to assign value to the element your using (like btnLeftElement, btnRightElement...) if you watch in your dev tools you will see the errors to correct. Keep working hard like you do!

PR #1388, opened 2021-06-10
Humberto WDFT-RMT-MAY21

humberto1212:


TA-Remote:
Hey Humberto!
Nice try on the lab. You were not that far from making the chrono work.
- You mixed the decimals and units in the index.js get seconds and get minutes functions. The units were [0] and decimals [1]
- In the index.js when you call the chronometer.start function you passed the function printime() as a parameter. You don't need to pass the () when you pass a function in a parameter. Just delete those () and you will see that your chrono works :) 
- I advise you go through the solution once it's up on the portal so you can check the splits and reset button too :) 
But nice overall!

PR #1390, opened 2021-06-10
partial solve

gigi-loranca:


TA-Remote:
Hey Guillermo, 
I just checked your lab. Nice try on the first iteration. Some part of your chronometer would work. 
But without anything in the index.js it's hard to tell. 
I suggest you check the solution once it's up and do the lab again just so you can see how the flow works and how you could have done it :)

PR #1391, opened 2021-06-10
Anja WDFT-RMT-MAY21

AnjaRL:
very tough one so far...
solved lab without bonus and some functions printSplit clearSplits setResetBtn

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,34 +1,48 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = null;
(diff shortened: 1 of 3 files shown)


TA-Remote:
Hey Anja! 
Very nice job on the lab. The logic is here so very well done. 
- You already saw with Manish why you needed an if condition in the start button so I won't go over it again. The rest of the methods are good though.
- Your code is clean so well done on that. Do not hesitate to comment more on the parts you could not solve. At least so we can see what you tried to do there :) 
- Your reset was not missing much. Check the solution when it's up to see what you could have done.
Have a nice day :)

PR #1402, opened 2021-07-07
[WDFT 286] Jessie

jessievp:
solved, but wasn't able to work through getting the MS. I may have repeated certain lines, making it redundant, but ommitting it fails the test cases. :(


jessievp:
reset doesn't function :( and the split time doesn't reflect :'(

PR #1409, opened 2021-07-12
[FT-RMT-072021] KarinaNarkeviciute

Karina-N:


Karina-N:
Time is still not being displayed on the clock.

PR #1412, opened 2021-07-13
complete Iteration 1

kpla001:


kpla001 (commented):
I leveraged Vadim's pull request and manually copied his code in order to better understand the lab, as well as to practice. The counter for milliseconds does not work, and I am not sure why not.

PR #1477, opened 2021-09-27
[FT-RMT-092021] Dominika Lipka

dominikalipka:


dominikalipka:
Iterations 2, 3 and 4 added.
I know that I should split my code between given functions but I realized that after finishing all the iterations and I was too tired to amend that... I thought that maybe somewhere later in the instructions there will be something said about them.
Code still works fine -> the chronometer is working! :)

PR #1483, opened 2021-09-28
[WD-FT 200921] Katharina Felgitsch

katiFe:


katiFe:
Hej Karina, 

I have a major problem with time management. By the time I am through the student portal its already running late. I get stressed, blocked and I don't see the trees in the forest anylonger. Again I would be super greatful if we could go through the lab again. Thank you

PR #1484, opened 2021-09-28
fix: 00:00:00 positions

abernier:


abernier:
@ross-u `0`s were wrongly positioned

ross-u (approved):

PR #1517, opened 2021-10-25
[FT-RMT-102021] Tatiana_Prada

TatianaPrada:
First interaction done.


TatianaPrada:
I have done all the interactions, however, the time is not printing well :/ 

I need some help with that issue.

PR #1582, opened 2021-11-20
Solved lab Rachouan

Rachouan:
Like I mentioned I just made the chronometer the calculator of the milliseconds so that we have 1 interval instead of multiple. Test will work if you change 10 to 1000 and remove all the 100s


carlarsmendes:
Yup, looks good :) Thanks! Nice job! This is a fun one to work on I think

PR #1584, opened 2021-11-20
WDPT Lis Nov21 Michael 

micfol:


micfol:
Milliseconds just displays 0 - and when I click reset, nothing happens, split list still remains, and clock won't reset to zero.

PR #1619, opened 2022-01-17
Nico Tollenaar - Part 1 done

NicoTollenaar:


javascript/chronometer.js
```diff
@@ -1,37 +1,67 @@
...
 
   start(callback) {
-    // ... your code goes here
+    this.intervalId = setInterval(() => {
+      if (callback) {
+        callback();
+      }
+      this.currentTime++;
+    }, 1000);
+    return this.currentTime;
```
TA-Remote, line 14: good job Nico, but it was not necessary to return this.currentTime in the start function. And if you don't return it, you could have used this.currentTime instead of this.start() later, e.g. this.currentTime % 60.

javascript/chronometer.js
```diff
@@ -1,37 +1,67 @@
...
+      return "00";
+    }
+    if (stringValue.length === 2) {
+      return stringValue;
+    } else if (stringValue.length === 1) {
+      paddedString = `0${stringValue}`;
+      return paddedString;
+    } else {
+      throw new Error("Something went wrong!");
+    }
```
TA-Remote, line 40: Basically a good solution and therefore you can leave it as it is. However, the solution is unnecessarily long and it would have been possible like this, for example:
```
computeTwoDigitNumber(number) {
    return ("0" + number).slice(-2);
  }
```
In this case I think that the long version is not necessarily more understandable

javascript/chronometer.js
NicoTollenaar: Thanks - this.start() give me a headache in part 2.

javascript/chronometer.js
NicoTollenaar: nice solution - shorter and clearer, I'll remember this one

javascript/index.js
```diff
@@ -14,52 +14,94 @@ const milUniElement = document.getElementById('milUni');
...
 btnRightElement.addEventListener('click', () => {
-  // ... your code goes here
+  if (btnRightElement.innerText === "SPLIT") {
+    printSplit();
+  }
+  if (btnRightElement.innerText === "RESET") {
+    clearSplits();
+    chronometer.reset();
+  }
 });
```
TA-Remote, line 105: Very good that you even made the bonus! 
I would have recommended you for the bonus rather to make another setInterval in the start function. One with 10 and one with 1000 milliseconds. I think that would have been easier for you as well. But there are as you have shown of course several possibilities.

PR #1622, opened 2022-01-17
[WDFT-Jan22] Day 6 Lab 2 - Doan Tran

Dorrrn:
Iteration 1 solved

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -1,34 +1,56 @@
 class Chronometer {
   constructor() {
     // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = null;
   }
 
   start(callback) {
     // ... your code goes here
+    this.intervalId = setInterval(() => {
+      return this.currentTime++;
+    }, 1000);
   }
 
   getMinutes() {
     // ... your code goes here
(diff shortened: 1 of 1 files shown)


javascript/chronometer.js
```diff
@@ -31,10 +31,10 @@ class Chronometer {
...
   }
 
-getMilSeconds(){
-  let milSeconds = this.currentime % 100;
-  return milSeconds;
-}
+  getMilSeconds() {
+    let milSeconds = (this.currentime * 1000) % 1000;
+    return milSeconds;
+  }
```
TA-Remote, line 37: With miliseconds it is a bit more complicated. Instead of converting it here from seconds, you might want to update your `start()` method. Either add another setInterval(), or update the existing setInterval() but increase the speed 

So for example every 10miliseconds you want to do something, be increasing the milisecond counter.

TA-Remote (commented):
Really great work!

A clue for split() functionality, what you need to do:
- create list element every time you click 'split'
- save result of the current time in it
- using appendChild, add this element so it appears on the page

javascript/chronometer.js
Dorrrn: thanks for reviewing and for the hints.. :)

PR #1625, opened 2022-01-17
WDFT GJ Day 6 Javascript Chronometer

gjschoutsen:
Chronometer Iteration 1 compleet.
Did not know why the split method would be handy. :(


TA-Remote:
When we start on further iterations you'll see why split() is useful!  we'll  have our chronometer running and will be able to print split time on the screen (for example when you are running a lap, you can take split time for each lap you run right :) )

PR #1629, opened 2022-01-17
Aleksandra chronometer lab

Aleksandrawoslx:


javascript/index.js
```diff
@@ -54,12 +68,34 @@ function setResetBtn() {
...
+    chronometer.start();
+  } else {
+    chronometer.stop();
+    btnLeftElement.innerText = 'START';
+    btnRightElement.innerText = 'RESET';
+    btnLeftElement.classList.toggle('stop');
+    btnLeftElement.classList.toggle('start');
+    chronometer.stop();
+  }
 });
```
TA-Remote, line 91: Here you should be extracting logic to separate functions -> code will be easier to read and follow. 
For example:

```
function setStartBtn() {
  btnLeftElement.classList.toggle('start');
  btnLeftElement.innerText = 'START';
}
```

and then call `setStartBtn()` on line 78.
So finally your event listener would have this. And logic is extracted in separate functions

```
btnLeftElement.addEventListener('click', () => {
  if(btnLeftElement.classList.contains('start')) {
    chronometer.start(printTime);
    setStopBtn();
    setSplitBtn();
  } else {
    chronometer.stop();
    setStartBtn();
    setResetBtn();
  }
});
```

PR #1662, opened 2022-01-19
done

throwoutofcoffeeexception:


javascript/chronometer.js
```diff
@@ -1,34 +1,44 @@
...
+    this.intervalId = null;
   }
 
   start(callback) {
-    // ... your code goes here
+    this.intervalId = setInterval(() => {
+      this.currentTime++;
+      callback();
+    }, 10)
   }
```
TA-Remote, line 12: Here you should be checking if the argument is passed into `start()` method. To be sure app does not crash or throw error

`if (callback) { callback() }`

TA-Remote (commented):
Excellent solution to the lab! 🙌  Unless you have any questions yourself

PR #1718, opened 2022-04-04
WDFTREM Mar2022 Mafalda

mafaldao21:


javascript/chronometer.js
```diff
@@ -49,9 +47,9 @@ class Chronometer {
   }
 
   split() {
-    let minutes= this.computeTwoDigitNumber(this.getMinutes())
-    let seconds= this.computeTwoDigitNumber(this.getSeconds())
-    return `${minutes}:${seconds}`;
+    let updatedMinutes= this.computeTwoDigitNumber(this.getMinutes(this.currentTime));
+    let updatedSeconds= this.computeTwoDigitNumber(this.getSeconds(this.currentTime));
```
TA-Remote, line 51: if you look at the getMinutes() method above, its not expecting any arguments, which means you should not be assing any when calling the function

TA-Remote (commented):
In 2nd iteration `this.currentTime` is already round number, so flooring it won't do anything. You should use `Math.floor()` on the final value before you return it.

Also not neccessary to check if currentTime is 0 because if it is , any math calculation would return zero anyways

PR #1721, opened 2022-04-04
Solved lab

lukecin:


javascript/chronometer.js
```diff
@@ -1,34 +1,57 @@
...
+    }, 1000)
   }
 
   getMinutes() {
-    // ... your code goes here
+    let minutes = 0;
+    if (this.currentTime >= 60) {
+      minutes++
+    }
+    return Math.round(this.currentTime / 60);
```
TA-Remote, line 21: Hi Dani, I think you saw that in class: just change Math.round to Math.floor:
`return Math.floor(this.currentTime / 60);`

javascript/chronometer.js
```diff
@@ -1,34 +1,57 @@
...
-    // ... your code goes here
+    let seconds = this.currentTime;
+    // if (this.currentTime % 60 === 0) {
+    //   return 0;
+    // }
+    // else 
+    if (this.currentTime <= 59) {
+      return seconds++;
+    }
   }
```
TA-Remote, line 33: here it's `return this.currentTime % 60`

PR #1724, opened 2022-04-04
[FTRMT032022] Oscar Gonzalez

oscargonzalezdev:
- [x] solve the lab javascript chronometer
- [x] pass all tests

### DIFF
--- a/javascript/chronometer.js
+++ b/javascript/chronometer.js
@@ -3,27 +3,27 @@ class Chronometer {
     this.currentTime = 0;
     this.intervalId = null;
   }
-
+  
   start(callback) {
     this.intervalId = setInterval(() => {
-      this.currentTime += 1;
+      this.currentTime++;
+      if (callback) {
+        callback();
+      }
     }, 1000);
-    if(callback){
-      setInterval((callback), 1000);
-    }
   }
 
   getMinutes() {
-    return Math.floor(this.currentTime/60);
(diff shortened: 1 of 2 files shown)


javascript/chronometer.js
```diff
@@ -1,34 +1,46 @@
...
 
   start(callback) {
-    // ... your code goes here
+    this.intervalId = setInterval(() => {
+      this.currentTime += 1;
+    }, 1000);
+    if(callback){
+      setInterval((callback), 1000);
+    }
   }
```
TA-Remote, line 14: Now you have two separate intervals running! Which is not exactly correct, we want to stick to just one interval, and check for callback inside it 

```
  start(callback) {
      this.currentTime++;
      if (callback) {
        callback();
      }
    }, 1000);
```

javascript/chronometer.js
```diff
@@ -1,34 +1,46 @@
...
 
   reset() {
-    // ... your code goes here
+    this.currentTime = 0;
   }
 
   split() {
-    // ... your code goes here
+    let formattedMinutes = this.computeTwoDigitNumber(this.getMinutes(this.currentTime));
+    let formattedSeconds = this.computeTwoDigitNumber(this.getSeconds(this.currentTime));
```
TA-Remote, line 42: If you look back at getMinutes() method above - it does not expect any arguments, so you do not want to be passing anything when you call it ;)

PR #1730, opened 2022-04-05
great exercise !!

sina-kaabi:


javascript/chronometer.js
```diff
@@ -1,10 +1,68 @@
...
+      }
+    }
+
+  
+  // The following is required to make unit tests work.
+  /* Environment setup. Do not modify the below code. */
+  if (typeof module !== 'undefined') {
+    module.exports = Chronometer;
   }
+
```
TA-Remote, line 49: Hi Sina! You need to delete everything after line 49 in chronometer.js.
You have every function twice. The tests for chronometer.js will not work if you have the functions etc. more than one time.
Again: tests are only for chronometer.js, we don't have tests for index.js

javascript/chronometer.js
```diff
@@ -1,10 +1,68 @@
...
     // ... your code goes here
+  
+   
+        this.currentTime = 0;
+        this.intervalId = null ;
+      }
+    
+      start(callback) {
+        this.intervalId=setInterval(()=> {
+         callback()
```
TA-Remote, line 13: here you need the condition:
```
if (callback) { 
callback()
}
```
after writing this if condition for the callback function, all tests are passing

javascript/index.js
```diff
@@ -40,26 +42,69 @@ function clearSplits() {
...
-btnLeftElement.addEventListener('click', () => {
+
   // ... your code goes here
-});
+  btnLeftElement.addEventListener('click', () => {
+
+  if(btnLeftElement.textContent === 'START' ) {
+    btnLeftElement.textContent = 'STOP';
+    btnRightElement.textContent = 'SPLIT';
+    chronometer.start(callback);
```
TA-Remote, line 73: the word "callback" is only a placeholder (in chronometer.js) so here you need to pass a function that you actually wrote (--> printTime)

javascript/index.js
```diff
@@ -40,26 +42,69 @@ function clearSplits() {
...
+  }
+  else {
+    btnRightElement.textContent = 'RESET';
+    btnLeftElement.textContent = 'START';
+    chronometer.split();
+
+  }
+
+
+
```
TA-Remote, line 106: I don't have much more to say since you didn't do that much.
We'll try to explain a few things in class later. The lab was hard, so you'd better go back and try to understand.

PR #1731, opened 2022-04-05
solved lab chronometter

ihab5520:


javascript/chronometer.js
```diff
@@ -1,34 +1,46 @@
...
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0
+    this.intervalId = null
   }
 
   start(callback) {
-    // ... your code goes here
-  }
+    this.intervalId = setInterval(()  =>{
```
Matteo-Ironhack, line 8: the only suggestion I can give is to work on your indentation.
this section would be more readable if it was written like this:
```
this.intervalId = setInterval(()  =>{
    if (callback)
        callback()
    this.currentTime++},1000)  }
```

PR #1732, opened 2022-04-05
solved lab

didicozaur:


javascript/chronometer.js
```diff
@@ -1,30 +1,44 @@
 class Chronometer {
-  constructor() {
-    // ... your code goes here
+  constructor(currentTime, intervalId) {
```
TA-Remote, line 2: Hi Diana, 
here it's just `constructor() {` without arguments

javascript/chronometer.js
```diff
@@ -1,30 +1,44 @@
...
-    // ... your code goes here
+    clearInterval(this.intervalId);
   }
 
   reset() {
-    // ... your code goes here
+    return (this.currentTime = 0);
   }
 
   split() {
```
TA-Remote, line 44: I think split should be clear after this morning, if not: please ask!

PR #1734, opened 2022-04-05
WDFT MARCH2022 Valentin Paul

Valentin-Paul:
in getMinutes I had to divide by 65 to get it to work.. I thought it would be 60 (60 sec = 1 min)
Didn't understand that

### DIFF
--- a/index.html
+++ b/index.html
@@ -17,11 +17,9 @@
 
     <section id="clock">
       <div id="sphere">
-        <span id="minDec" class="number">0</span>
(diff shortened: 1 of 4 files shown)


TA-Remote (commented):
Good work on the lab! 

Interesting way to solve the display of time, having one string rather than string for each digit. Make sure you adjust css accordingly right ;)

Also always pay attention if the function/method is expecting an argument -> if it does not - do not pass anything when you call it.

We will review how to split time during lecture

PR #1735, opened 2022-04-05
solved

infjay:


javascript/chronometer.js
```diff
@@ -1,34 +1,40 @@
...
-
+ 
   getSeconds() {
-    // ... your code goes here
+  return Math.floor(this.currentTime % 60)
   }
 
   computeTwoDigitNumber(value) {
     // ... your code goes here
   }
```
TA-Remote, line 26: Hi Jawid! One way to solve this one would be like this:
 
    return ("0" + value).slice(-2);
  }
  
it will always add 0 and return only the last two values. so if you have 10 --> 010 --> slice(-2) --> 10, if that makes sense or 9 --> 09 --> slice(-2) --> 09.

PR #1737, opened 2022-04-07
change readme instruction for Print our chronometer

Joanneseiler:
We wanted to send you some thoughts on the lab instructions.
As you know the Chronometer Lab is one of the trickiest challenges in Module 1 and the instructions don't make it any easier:

Some iterations are complex & description is very very long (for example, iteration 2).

Suggestions:
- split lab into smaller steps, each of them with a clear goal (example. split iteration 2 in two steps, "Changing buttons" + "Print Time")
- improve instructions for each step

I deleted this line -> „Using our `Chronometer` methods to get the values, this should be easy!“ and added something else. 

That's all we've thought about so far. If you have time to look at it, it would certainly be helpful for the upcoming students.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -226,7 +226,9 @@ We will be working on the `javascript/index.js` file. We need to do the followin
 
 Each second we need to update our screen. So go ahead and create a function that will receive the value for minutes and seconds, and print that on our HTML.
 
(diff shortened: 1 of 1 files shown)


Joanneseiler:
Update: Marian changed instructions to make life easier for you and students!

PR #1748, opened 2022-04-18
LISWBFT0422 MARCELO MILHOMEM Done

marcelomilhomem:


javascript/chronometer.js
```diff
@@ -1,33 +1,51 @@
 class Chronometer {
   constructor() {
+    this.currentTime = 0;
+    this.intervalId = null;
     // ... your code goes here
   }
 
   start(callback) {
```
webLisbonTA, line 8: You missed to dd the callback logic. It is said that the callback is a function that can be optionally passed as an argument. So the idea was to check if there is a callback passed, if yes it should be executed inside the setInterval. Like this:

```
  start(callback) {
      this.currentTime++;
      if (callback) {
        callback();
      }
    }, 1000);
  }

```

webLisbonTA (commented):
Good job, Marcelo

webLisbonTA (commented):
Great job on finishing the lab!

PR #1755, opened 2022-04-18
[WDFT042022] Marina Palma

MarinaPalma:


javascript/chronometer.js
```diff
@@ -1,34 +1,43 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = null;
   }
 
   start(callback) {
```
webLisbonTA, line 7: You missed to check add the callback logic. It is said that the callback is a function that can be optionally be passed as an argument. So the idea was to check if there is a callback passed, if yes it should be executed inside the setInterval. Like this:
```
  start(callback) {
      this.currentTime++;
      if (callback) {
        callback();
      }
    }, 1000);
  }

```

webLisbonTA (commented):
Good job Marina!

webLisbonTA (commented):
Great job on finishing the lab!

PR #1759, opened 2022-04-18
[FTAPR2022] João Morgado

JoaoMorgado8:


javascript/chronometer.js
```diff
@@ -1,34 +1,46 @@
...
   }
 
   getMinutes() {
-    // ... your code goes here
+    return Math.floor(this.currentTime / 60);
   }
 
   getSeconds() {
-    // ... your code goes here
+    return Math.floor(this.currentTime % 60);
```
webLisbonTA, line 21: Modulo already gives you a full integer, so you don't need the Math.floor()

webLisbonTA (commented):
Great job!

PR #1761, opened 2022-04-19
[FTLIS0422] Duarte Alves

drttyy:


javascript/chronometer.js
```diff
@@ -1,34 +1,46 @@
...
   }
 
   getMinutes() {
-    // ... your code goes here
+    return Math.floor(this.currentTime / 60);
   }
 
   getSeconds() {
-    // ... your code goes here
+    return Math.floor(this.currentTime % 60);
```
webLisbonTA, line 21: Modulo already gives you a full integer, so there is no need for Math.floor()

javascript/chronometer.js
```diff
@@ -1,34 +1,46 @@
...
   }
 
   computeTwoDigitNumber(value) {
-    // ... your code goes here
+    if (value < 9) {
```
webLisbonTA, line 25: You are not including 9, < 9 does not include it. You should do : <= 9 or < 10

webLisbonTA (commented):
Great job!

webLisbonTA (commented):
Nice work finishing the lab. Your split/reset button is not working properly, it gives the splits but it resets when I do it, you used some functions in the wrong place but you have all the logic there

PR #1769, opened 2022-04-25
done with little problems

EvgeniiaRiaguzova:


javascript/chronometer.js
```diff
@@ -1,34 +1,40 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime=0;
+    this.intervalId=null;
   }
 
   start(callback) {
-    // ... your code goes here
+    const intervalId = setInterval(() => {this.currentTime +=1}, 1000);
```
Matteo-Ironhack, line 8: you should have also checked for callback to be defined and executed callback. other than that, good use of an arrow function to preserve the `this` context and access instance's value

javascript/chronometer.js
```diff
@@ -1,34 +1,40 @@
...
 
   getSeconds() {
-    // ... your code goes here
+   let numberOfSeconds = this.currentTime % 60
+   return numberOfSeconds
   }
 
   computeTwoDigitNumber(value) {
-    // ... your code goes here
+    let twoDigitNumber
```
Matteo-Ironhack, line 22: if you are not using a variable, you probably should not declare it. Otherwise, clean solution

javascript/chronometer.js
Matteo-Ironhack: also, the fact that you do not assign the interval's return value (intervalId) to the instance (by using `this.intervalId`, makes other tests fail (for example the stop function)

javascript/chronometer.js
```diff
@@ -1,34 +1,40 @@
...
   }
 
   reset() {
-    // ... your code goes here
+    this.currentTime=0;
   }
 
   split() {
-    // ... your code goes here
+    return `${numberOfMinutes}:${numberOfSeconds}`
```
Matteo-Ironhack, line 37: these two variables are not defined. this section is doable using only the class' methods (getMinutes, getSeconds and split). Take a look at the solution to see a way of doing it

PR #1773, opened 2022-04-25
lab done

ghost:


javascript/chronometer.js
```diff
@@ -1,34 +1,53 @@
...
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = null;
   }
 
   start(callback) {
-    // ... your code goes here
-  }
+  
+    this intervalID = setInterval(() => {
```
Matteo-Ironhack, line 9: should be `this.intervalId`; careful to not forget dots

javascript/chronometer.js
```diff
@@ -1,34 +1,53 @@
...
+      this.currentTime = intervalCounter += 1;
+      
+      if(callback){
+      return callback;
```
Matteo-Ironhack, line 13: this should not return the callback, but execute it (like `callback()`

javascript/chronometer.js
```diff
@@ -1,34 +1,53 @@
...
+      }
+
+    }, 1000);
+   
+   
+    }
+}
 
   getMinutes() {
```
Matteo-Ironhack, line 22: all the methods under line 21 are out of the class. the curly brace in line 20 closes the class

javascript/chronometer.js
```diff
@@ -1,34 +1,53 @@
...
-    // ... your code goes here
+  
+    let Minutes = Math.floor this.currentTime /60;
```
Matteo-Ironhack, line 24: Math.floor is function and thus should have parenthesis to call it: this line would be `Math.floor(this.currentTime /60)`

javascript/chronometer.js
```diff
@@ -1,34 +1,53 @@
...
+    return getMinutes;
```
Matteo-Ironhack, line 25: what happens if you do this is that you return the function you are executing. You should return Minutes to get the expected return value

javascript/chronometer.js
```diff
@@ -1,34 +1,53 @@
...
+
   }
 
   getSeconds() {
-    // ... your code goes here
-  }
+   
+    let numberofSeconds = this.currentTime % 60;
+    return getSeconds;
```
Matteo-Ironhack, line 32: same thing here as with getMinutes

javascript/chronometer.js
```diff
@@ -1,34 +1,53 @@
...
 
   computeTwoDigitNumber(value) {
-    // ... your code goes here
+    if(value < 10 ) return 0
+    return value;
   }
 
   stop() {
-    // ... your code goes here
+    clearInterval this.intervalId;
```
Matteo-Ironhack, line 41: clearInterval, like Math.floor, is a function. In order to run it you have to use parenthesises.

PR #1777, opened 2022-04-26
solved lab

nataliabuzila:


javascript/chronometer.js
```diff
@@ -1,34 +1,50 @@
 class Chronometer {
   constructor() {
-    // ... your code goes here
+    this.currentTime = 0;
+    this.intervalId = null;
   }
 
   start(callback) {
-    // ... your code goes here
+    if(callback) {
```
Matteo-Ironhack, line 8: what happens here if I don't have a callback function? Your program should work regardless (if the callback argument is defined, execute it; if not, then continue like usual)

javascript/chronometer.js
```diff
@@ -1,34 +1,50 @@
...
+      TwoDigitNumber = "0"+ value;
+    } else if (value<99) {
+      TwoDigitNumber = value.toString();
+    }
+  return TwoDigitNumber;
   }
 
   stop() {
-    // ... your code goes here
+    clearInterval(this.intervalId);
```
Matteo-Ironhack, line 37: the problem mentioned in your start functions makes this function not work, but the code is otherwise valid

